PR #857, opened 2014-11-10
Update editor only once after execution of save actions

sschaef:
Updating the editor is expensive, therefore we only want to do it a
single time - at the end of the save action running phase.

Fixes #1002309

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/SaveActionExtensions.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/SaveActionExtensions.scala
@@ -12,6 +12,7 @@ import org.eclipse.jdt.core.ICompilationUnit
 import org.eclipse.jdt.internal.ui.javaeditor.saveparticipant.IPostSaveListener
 import org.eclipse.jface.text.IDocument
 import org.eclipse.jface.text.IRegion
+import org.eclipse.jface.text.ITextSelection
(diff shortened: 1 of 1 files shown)


org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/SaveActionExtensions.scala
```diff
@@ -102,18 +120,39 @@ trait SaveActionExtensions extends HasLogger {
...
-    undoManager.endCompoundChange()
+      val undoManager = DocumentUndoManagerRegistry.getDocumentUndoManager(udoc)
+      undoManager.beginCompoundChange()
+
+      try applyDocumentExtensions(udoc)
+      catch {
+        // thrown when save actions had to be aborted
+        case e: IllegalStateException =>
+          updateEditor()
+          throw e
```
dragos: Is there no need to call `endCompoundChange()` when an exception is thrown? Or should we call it inside a `finally` block?

sschaef: Indeed, that is a bug. I'll wrap the entire code block in a finally clause.

dragos:
Anyone tried out this PR? I won't have time for code reviews in the next days, so any help is appreciated.

dragos:
LGTM! Thanks @sschaef

PR #868, opened 2014-11-26
Macro expansion support.

lnxmnx:
Macro functionality is done through Annotations. When clicking on
expand macro annotation macro expands into the source code, annotation
to collapse is added to expanded region. When clicking on collapse
macro annotation macro is erased with expandee.

When triggering save command performSave function of
ScalaSourceFileEditor is called. Before actual save every macro is
erased with its expandee, so that source file doesn't contain any
macro applications.

Through createLineNumberRulerColumn function line number
column for custom line numbering is integrated. It internally stores regions, where
macros are and change line numbers where neccesary.

Macros can be turned off by clicking on the Enable macro icon,
M1+M2+M(Ctrl-Shift-M) shortcut or through preference page. When turned off
all macros are erased from sources.


Andrei-Pozolotin:
please merge so we can try it out in nightly builds

sschaef:
It doesn't work correctly therefore no merge. @lnxmnx I'm closing this. Please reopen when you have a new version.

dragos:
..but it would still be great to have this in once the rough edges are gone!

dragos:
@Andrei-Pozolotin you can find binaries of this PR at this link:

http://download.scala-ide.org/pr-builds/scala-ide/pr-868/2.11.x/site

(same is true for any pull request)

Andrei-Pozolotin:
@dragos thanks, good to know, checking

PR #870, opened 2014-11-30
Add strip margins in string interpolation 

sschaef:
This did not work because the start and end position of the partition
where the cursor is placed was used. This is a problem because string
interpolators are split up into multiple partitions - therefore we have
to traverse the partitions until we found the start and end of the multi
line string literal.

Note that this behavior is still buggy in case someone inserts a multi
line string literal inside of a code block that is already part of a
string interpolation:

    s"""${
      val s = """str"""
      s
    }"""

Fortunately, such code blocks aren't used often, therefore I consider
this as a won't fix until someone opens a ticket..

Fixes #1002145


dragos:
LGTM.

Thanks, @sschaef this is indeed very useful!

PR #880, opened 2014-12-17
Fix double-newline bug in Scala templates

Redattack34:
This should fix ticket #1002303. I've also added some basic tests for the ScalaTemplateContext class since it didn't have any before.


sschaef:
I forgot this PR, sorry. LGTM and thanks for the fix.

PR #887, opened 2015-01-01
Do not pass DocDef with underlying ClassDef to Namers.standardEnterSym

sschaef:
`standardEnterSym` internally uses the following code:

    tree.symbol match {
      case NoSymbol => try dispatch() catch typeErrorHandler(tree, this.context)
      case sym      => enterExistingSym(sym, tree)
    }

where tree is a `DocDef` with an underlying `ClassDef`. Because `symbol`
on `DocDef` is forwarded to the underlying tree, we would pass a symbol
that doesn't match its original tree to `enterExistingSym`, which later
would result in an `ClassCastException`.

Fixes #1002367

---
I'm not sure if this is the right location to fix this bug. Should it be done in `Namers.standardEnterSym` directly?

Review by @huitseeker would be nice.


dragos:
Thanks @sschaef for looking at this, but in the meantime this was fixed upstream, in [PR 4205](https://github.com/scala/scala/pull/4205) and should be part of 2.11.5.

PR #891, opened 2015-01-17
Copy qualified name addendum

mlangc:
This pull request contains some minor improvements to "Copy Qualified Name", lot's of additional unit tests and improved test APIs.

### DIFF
--- a/org.scala-ide.sdt.core.tests/src/org/scalaide/core/compiler/NamePrinterTest.scala
+++ b/org.scala-ide.sdt.core.tests/src/org/scalaide/core/compiler/NamePrinterTest.scala
@@ -387,6 +387,121 @@ class NamePrinterTest {
       "WithLazyVal.x")
   }
 
+  @Test
+  def testWithMethodCallInNestedStructure() {
+    testWith(
+      """|object MethodcallInNestedStructure {
+         |  def method = 999
+         |}
+         |class MethodcallInNestedStructure {
+         |  def otherMethod {
+         |     MethodcallInNestedStructure.method/**/
+         |  }
+         |}""",
+      "MethodcallInNestedStructure.method")
+  }
+
+  @Test
+  def testWithNamedParamCtorCallInMethod() {
+    testWith(
+      """|package test.named.param.ctor.call
+         |class NamedParamCtorCallInMethod(arg: String) {
+         |  def method {
+         |    new NamedParamCtorCallInMethod(arg/**/ = "aaahrg")
+         |  }
+         |}""",
+      "test.named.param.ctor.call.NamedParamCtorCallInMethod(arg: String)")
+  }
+
+  @Test
+  def testWithSimpleAnnotation() {
+    testWith(
+      """|import scala.annotation._
+         |class SimpleAnnotation extends StaticAnnotation
+         |@SimpleAnnotation/**/
+         |class WithSimpleAnnotation
+      """,
+      "SimpleAnnotation")
+  }
+
+  @Test
+  def testWithAnnotationOnDefOnParam() {
+    testWith(
+      """|import scala.annotation._
+         |class AnnotationOnDefOnParam(val name: String) extends StaticAnnotation
+         |object AnnotationOnDefOnParam {
+         |  @AnnotationOnDefOnParam(name/**/ = "Hallo")
+         |  def test = 32
+         |}""",
+      "AnnotationOnDefOnParam(name: String)")
+  }
+
+  @Test
+  def testWithAnnotationOnDef() {
+    testWith(
+      """|import scala.annotation._
+         |class AnnotationOnDef(val name: String) extends StaticAnnotation
+         |object AnnotationOnDef {
+         |  @AnnotationOnDef/**/(name = "Hallo")
+         |  def test = 32
+         |}""",
+      "AnnotationOnDef")
+  }
+
+  @Test
+  def testWithJavaxGeneratedAnnotationOnClass() {
+    testWith(
+      """|import javax.annotation._
+         |@Generated/**/(Array("today"))
+         |class WithJavaxGeneratedAnnotationOnClass""",
+      "javax.annotation.Generated")
+  }
+
+  @Test
+  def testWithJavaxGeneratedAnnotationOnMethod() {
+    testWith(
+      """|import javax.annotation._
+         |class WithJavaxGeneratedAnnotationOnMethod {
+         |  @Generated/**/(Array("yesterday"))
+         |  def method = Unit
+         |}""",
+      "javax.annotation.Generated")
+  }
+
+  @Test
+  def testWithJavaxResourceAnnotationOnMethod() {
+    testWith(
+      """|import javax.annotation._
+         |class WithJavaxResourceAnnotationOnResource {
+         |  @Resource/**/(name = "test", lookup = "anywhere")
+         |  def method = Unit
+         |}""",
+      "javax.annotation.Resource")
+  }
+
+  @Test
+  def testWithAxillaryCtor() {
+    testWith(
+      """|class WithAxillaryCtor(str: String) {
+         |  def this(i: Int) = this(i.toString)
+         |  def foo() { new WithAxillaryCtor(/**/33) }
+         |}""",
+      "WithAxillaryCtor(i: Int)")
+  }
+
+  @Test
+  def testWithMethodCallOnGeneric() {
+    testWith(
+      """|class WithMethodCallOnGeneric[A](a: A) {
+         |  def method[B, C](b: B, c: C) = b.hashCode + c.hashCode
+         |  def foo = {
+         |    val tst = new WithMethodCallOnGeneric("string")
+         |    tst.method(/**/1L, 0.0)
+         |  }
+         |}""",
+      "WithMethodCallOnGeneric.method[B, C](b: B, c: C)")
+  }
+
   private def testQnameWith(input: String, expected: Option[String]) {
     val source = input.stripMargin
     val cu = mkScalaCompilationUnit(source)
--- a/org.scala-ide.sdt.core.tests/src/org/scalaide/core/compiler/NamePrinterTest.scala
+++ b/org.scala-ide.sdt.core.tests/src/org/scalaide/core/compiler/NamePrinterTest.scala
@@ -480,7 +480,7 @@ class NamePrinterTest {
   }
 
   @Test
-  def testWithAxillaryCtor() {
+  def testWithAuxillaryCtor() {
     testWith(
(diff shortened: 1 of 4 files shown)


mlangc:
```
ssh_exchange_identification: Connection closed by remote host
scp: scala-ide.dreamhosters.com/pr-builds/scala-ide/pr-891/2.11.x: No such file or directory
Build step 'Execute shell' marked build as failure
Run condition [Current build status] enabling perform for step [Publish JUnit test result report]
Recording test results
Finished: FAILURE
```
Looks like the build infrastructure has a problem...

dragos:
retest this please

dragos:
@mlangc as the name printer gets more and more complex, how does it compare to the `DeclarationPrinter`? Do you think they could merge? I think the initial impression that it's easier to rewrite it might change once you get a grip on the full structure of Scala types. I still think we should have only one composable declaration printer, based on either of the two.

mlangc:
@dragos I've not forgotten about your comment; it's just so that I'm away during the weekend. I'll answer your questions in a few days.

dragos:
Cool! It's a general question, we can discuss on the mailing list as well. It shouldn't block this PR anyway

mlangc:
Very well then.. is there anything else I should address?

org.scala-ide.sdt.core.tests/src/org/scalaide/core/compiler/NamePrinterTest.scala
```diff
@@ -387,6 +387,121 @@ class NamePrinterTest {
...
+         |  @Resource/**/(name = "test", lookup = "anywhere")
+         |  def method = Unit
+         |}""",
+      "javax.annotation.Resource")
+  }
+
+  @Test
+  def testWithAxillaryCtor() {
+    testWith(
+      """|class WithAxillaryCtor(str: String) {
```
sschaef: `s/Ax/Aux` :)

org.scala-ide.sdt.core.tests/src/org/scalaide/core/compiler/NamePrinterTest.scala
```diff
@@ -387,6 +387,121 @@ class NamePrinterTest {
...
+  def testWithMethodCallOnGeneric() {
+    testWith(
+      """|class WithMethodCallOnGeneric[A](a: A) {
+         |  def method[B, C](b: B, c: C) = b.hashCode + c.hashCode
+         |  def foo = {
+         |    val tst = new WithMethodCallOnGeneric("string")
+         |    tst.method(/**/1L, 0.0)
+         |  }
+         |}""",
+      "WithMethodCallOnGeneric.method[B, C](b: B, c: C)")
```
sschaef: This functionality doesn't work for me when I use it in the UI, it only prints `WithMethodCallOnGeneric.method`. Does it work for you?

mlangc: You have to place the cursor on "(", like in the test; of course it would be nicer if this didn't make a difference, but I decided that it is not worth the effort to change this if I remember correctly. What do you think?

sschaef: Did you find out why it only works on the paren? And why is this not documented in the commit message/Scaladoc?

sschaef:
The tests are still unreliable. For now the only fix I see is to add a package declaration to all of them. Enumerate them or better use the name of the test to make it unique.

I get the following exception:
```
[testfile.scala]: exception during background compile: java.lang.NullPointerException: underlying.getLocation == null for: L/NamePrinterTest$/src/testpackage42012760892553/testfile.scala
java.lang.NullPointerException: underlying.getLocation == null for: L/NamePrinterTest$/src/testpackage42012760892553/testfile.scala
	at org.scalaide.core.resources.EclipseResource.path(EclipseFile.scala:77)
	at scala.reflect.internal.util.BatchSourceFile.hashCode(SourceFile.scala:189)
	at scala.runtime.ScalaRunTime$.hash(ScalaRunTime.scala:206)
	at scala.collection.mutable.HashTable$HashUtils$class.elemHashCode(HashTable.scala:409)
	at scala.collection.mutable.HashMap.elemHashCode(HashMap.scala:40)
	at scala.collection.mutable.HashTable$class.findEntry(HashTable.scala:132)
	at scala.collection.mutable.HashMap.findEntry(HashMap.scala:40)
	at scala.collection.mutable.HashMap.apply(HashMap.scala:64)
	at scala.tools.nsc.interactive.Global$$anonfun$backgroundCompile$5$$anonfun$apply$8.apply(Global.scala:586)
	at scala.tools.nsc.interactive.Global$$anonfun$backgroundCompile$5.apply(Global.scala:579)
	at scala.tools.nsc.interactive.Global$$anonfun$backgroundCompile$5.apply(Global.scala:579)
	at scala.collection.TraversableLike$WithFilter$$anonfun$foreach$1.apply(TraversableLike.scala:778)
	at scala.collection.immutable.List.foreach(List.scala:381)
	at scala.collection.TraversableLike$WithFilter.foreach(TraversableLike.scala:777)
	at scala.tools.nsc.interactive.Global.backgroundCompile(Global.scala:579)
	at scala.tools.nsc.interactive.PresentationCompilerThread.run(PresentationCompilerThread.scala:25)
```
@dragos Do you know what is the reason for this? Why is the compiler still compiling? When I add
```scala
    cu.scalaProject.presentationCompiler { compiler =>
      compiler.askReload(List(cu)).get
      val src = cu.sourceMap(source.toCharArray()).sourceFile
      compiler.askLoadedTyped(src, true).get
    }
```
after the `mkScalaCompilationUnit` call, to load them early, some of the tests even fail reliable (probably because they lack package declarations).

sschaef:
@mlangc Please replace the `assertEquals` in `testQnameWith` with `if (expected != actual) throw new ComparisionFailure("", expected, actual)`. This gives us better UI support in Eclipse.

sschaef:
Adding package declarations didn't fix the problems. The class names are already unique anyway.

dragos:
@sschaef I don't know what's happening there... are they always failing with an NPE?

org.scala-ide.sdt.core.tests/src/org/scalaide/core/compiler/NamePrinterTest.scala
dragos: Maybe because the tree is different from the case of simple method application. A generic call is

```
Apply(
  TypeApply(Select(..., "method"), List(B, C)))
  List(b, c))
```

org.scala-ide.sdt.core/src/org/scalaide/core/compiler/NamePrinter.scala
```diff
@@ -191,6 +218,7 @@ class NamePrinter(cu: InteractiveCompilationUnit) {
         case comp.Import(tree, selectors) => handleImport(loc, tree, selectors)
         case ident: comp.Ident => handleIdent(ident)
         case packageDef: comp.PackageDef => handlePackageDef(packageDef)
+        case apply: comp.Apply => handleApply(apply)
```
dragos: You might add here `TypeApply`, or maybe a supertype of the two, like `GenericApply`

dragos:
I am a bit confused by another thing: it's "copy qualified name", but it copies also the signature... how do people usually use this feature?

mlangc:
@sschaef @dragos See https://www.assembla.com/spaces/scala-ide/tickets/1002373-nameprintertest-is-flaky#/activity/ticket: for the flaky tests. I think this is a bug in the presentation compiler.

mlangc:
@dragos As for "copy qualified name" vs. ""copy qualified name with signature": I've looked at the corresponding feature for Java and mostly copied it's behavior.

org.scala-ide.sdt.core.tests/src/org/scalaide/core/compiler/NamePrinterTest.scala
mlangc: @sschaef I didn't document that because I thought that nobody will ever care. Generally I would appreciate it if we don't loose ourself in details that are of no concern to users. We've tons of bugs to fix out there that definitely are!

sschaef: Writing down your thoughts in the commit message is not a unnecessary detail to me - in fact I would be very happy if previous contributors to the IDE would have done it. In future, someone may care about this detail and ask themselves why it works the way it works. When there is a commit message - even if it is just a single line saying "arbitrary behavior without much concern about correctness" - it makes their job a lot easier. Writing such commit messages or Scaladoc comments is not a big deal when you do the commit - it just grows in size when you have to do it afterwards when reviewers have to ask questions about why the code works the way it works.

For now it is ok, but in future it would be nice if you would pay more attention to the commit messages+Scaladoc of your code.

sschaef:
On 01/29/2015 06:22 PM, Iulian Dragos wrote:
>
> @sschaef <https://github.com/sschaef> I don't know what's happening 
> there... are they always failing with an NPE?
>
In fact they don't even fail, the NPEs are thrown after the tests are 
already finished (which means that the project that contains their 
sources gets deleted).

org.scala-ide.sdt.core.tests/src/org/scalaide/core/compiler/NamePrinterTest.scala
mlangc: Thanks for your remarks, I'll take that into account in the future. Maybe you might want to take a look at http://www.amazon.de/Clean-Code-Handbook-Software-Craftsmanship/dp/0132350882. While I don't agree with everything that is written there, the chapter about comments really changed the way I was thinking about this matter.

dragos: @mlangc I agree we should be practical, there's a ton of work that needs to be done. Let's reach a decision on this (tests are already flaky, which is a pity, but I hope we can nail it down eventually) and merge.

dragos:
retest this please

dragos:
retest this please

dragos:
retest this please

dragos:
ok to test

dragos:
I have a fix for flaky test at dragos/scala-ide@2eb1def9a4

dragos:
Given that #899 builds upon this one, and brings much-needed fixes for flaky tests, I'd like to close this one. @sschaef are you happy with the state of this PR, so we can move to #899 and look only at the additional commit(s)?

sschaef:
Yes, let's move on to #899.

PR #892, opened 2015-01-23
Bump version to new Scala 2.11.6 development version, including xml and

dragos:
parser combinator dependencies.


dragos:
@skyluc if you have a moment, please take a look. I bumped generously the Manifest version dependency for Scala modules, since I missed it in the first place. Given Scala-flavored semantic versioning, I think that's ok.

skyluc:
LGTM

PR #893, opened 2015-01-25
Check return value of getResource call to avoid NPE

sschaef:
While the exception doesn't mention which subexpression throws the NPE,
it seems that only getResource can return null, at least that is what
its Javadoc says.

Fixes #1002375


dragos:
Thanks @sschaef ! LGTM

dragos:
Do you think this would make sense to be included in a 4.0.1 bugfix release? Or leave it for 4.1.0?

sschaef:
I'm not sure if such an error is shown by a notification dialog. If not, I think it can wait for 4.1. On Assembla, I asked the user who reported the bug how they found out about it.

PR #894, opened 2015-01-28
Correct debug messages in ScalaPlugin

sschaef:


dragos:
LGTM. The bot is down for maintenance, but I think this is safe to merge.

PR #896, opened 2015-02-01
Handle auto indent correctly after unicode representation of =>

sschaef:
No tests provided because no test suite exists. The unicode
representation may be used in pattern matching, function literals and
self references. For all of them a new indentation level needs to be
introduced.

Fixes #1002383


dragos:
LGTM

PR #897, opened 2015-02-02
Handle partially applied functions properly

mlangc:
Fix 1002381

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/decorators/bynameparams/CallByNameParamAtCreationPresenter.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/decorators/bynameparams/CallByNameParamAtCreationPresenter.scala
@@ -48,14 +48,36 @@ object CallByNameParamAtCreationPresenter extends HasLogger {
           super.traverse(tree)
         }
 
+        def isSynthetic(tree: compiler.Tree) = {
+          Option(tree.symbol).exists(_.isSynthetic)
+        }
+
+        def isByNameParam(param: compiler.Symbol) = {
+          param.isByNameParam || referencesByNameParam(param)
+        }
+
+        /*
+         * This should cover partially applied functions referencing by-name-params (see #1002381).
+         */
+        def referencesByNameParam(param: compiler.Symbol) = {
+          if (!param.hasRawInfo) {
+            false
+          } else {
+            param.rawInfo.typeSymbol match {
+              case cs: compiler.ClassSymbol => cs.name == compiler.tpnme.BYNAME_PARAM_CLASS_NAME
+              case _ => false
+            }
+          }
+        }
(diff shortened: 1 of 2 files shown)


org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/decorators/bynameparams/CallByNameParamAtCreationPresenter.scala
```diff
@@ -48,14 +48,36 @@ object CallByNameParamAtCreationPresenter extends HasLogger {
           super.traverse(tree)
         }
 
+        def isSynthetic(tree: compiler.Tree) = {
+          Option(tree.symbol).map(_.isSynthetic).getOrElse(false)
```
sschaef: can be simplified with `Option.exists`

sschaef:
Beside from the nitpick, LGTM

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/decorators/bynameparams/CallByNameParamAtCreationPresenter.scala
mlangc: Thanks for pointing this out!

sschaef:
Thanks!

dragos:
Not sure if introduced by this PR, but a name printer test failed (probably spurious failure):

https://jenkins.scala-ide.org:8496/jenkins/job/parameterized-scala-ide/1960/

Can you please have a look?

sschaef:
I guess the failure is due to the same reasons as discussed in #891.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/decorators/bynameparams/CallByNameParamAtCreationPresenter.scala
```diff
@@ -48,14 +48,36 @@ object CallByNameParamAtCreationPresenter extends HasLogger {
...
+        }
+
+        /*
+         * This should cover partially applied functions referencing by-name-params (see #1002381).
+         */
+        def referencesByNameParam(param: compiler.Symbol) = {
+          if (!param.hasRawInfo) {
+            false
+          } else {
+            param.rawInfo.typeSymbol match {
```
dragos: Why `rawInfo` instead of `info`?

mlangc: @dragos There is no specific reason for that... the tests also pass if you replace `rawInfo` with `info`, which should return the same result anyway if I read the sources correctly. Should I create another pull request for that, or do you just want to change that yourself?

PR #898, opened 2015-02-10
No need to run `FreshRunReq`, this is normal operation for the PC.

dragos:


dragos:
retest this please

org.scala-ide.sdt.core/src/org/scalaide/core/compiler/IScalaPresentationCompiler.scala
```diff
@@ -367,6 +367,10 @@ object IScalaPresentationCompiler extends HasLogger {
                 Thread.currentThread().interrupt()
                 None
 
+              case Right(f: FreshRunReq)          =>
+                logger.info("FreshRunReq in ask:\n" + f)
```
sschaef: `info` has an overload, which takes in its second argument a throwable. Please choose that one.

dragos: Note that `FreshRunRequest` has no stack trace, but I can use the overload nonetheless

sschaef:
Generally, LGTM. How did you find out about this issue? Was there a bug somewhere?

dragos:
I started to see entries in the Eclipse log (there's a catch-all at the end, who'd route the exception to the Eclipse error log).

PR #900, opened 2015-02-12
Fix NPE in semantic highlighting job

sschaef:
In case the editor is closed, no interactive compilation unit can be
retrieved and null is returned. This bug occurred while stepping through
multiple files in the debugger.

Fixes #1002386


dragos:
LGTM! thanks!

PR #901, opened 2015-02-23
T1002387 Remove an option to close SPCs when editors are open

mpociecha:
There are known cases when this (by default disabled) option leads to
memory leaks. Since the way in which dependencies of the presentation
compiler are freed hasn't been corrected so far, it's better to remove
this option. It can be brought back when the related problems will be
solved and someone will still consider it as something helpful.

This commit removes the preference stored previously in the preference
store, the checkox from GUI and underlying logic. It also updates
tests.


sschaef:
ok to test

sschaef:
LGTM

sschaef:
I completely oversaw this: To automatically close a ticket on Assembla mention it with `Fix #xxx`, don't put the ticket number in the title.

mpociecha:
Ok, thanks. I even didn't know there's such a feature. On the other hand, personally I consider it as something unfinished because an update to docs is not merged.

sschaef:
I don't have merge rights to the docs repo. @dragos will hopefully merge soon.

PR #903, opened 2015-02-26
Add command extension to restart presentation compiler with shortcut

sschaef:
Once the user configures a shortcut for the restart operation, it is
possible to restart the PC of the project that belongs to the active
selection.

No default shortcut provided because it is difficult to find one that is
not already used.

Fixes #1002391

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/ui/internal/actions/AbstractPopupAction.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/ui/internal/actions/AbstractPopupAction.scala
@@ -1,19 +1,37 @@
 package org.scalaide.ui.internal.actions
 
+import org.eclipse.core.commands.AbstractHandler
+import org.eclipse.core.commands.ExecutionEvent
 import org.eclipse.core.resources.IProject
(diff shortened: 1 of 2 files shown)


dragos:
LGTM! Thanks @sschaef for the speedy fix!

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/actions/AbstractPopupAction.scala
```diff
@@ -1,20 +1,38 @@
...
   private var selectionOption: Option[ISelection] = None
 
+  /**
+   * This method is called if either the menu entry is invoked or the handler
+   * (if it exists) is called when an [[IResource]] element is selected.
+   */
   def performAction(project: IProject)
 
+  override def execute(event: ExecutionEvent): AnyRef = {
+    EditorUtils.resourceOfActiveEditor flatMap (r ⇒ Option(r.getProject)) foreach performAction
```
dragos: I'm not a fan of unicode characters like this one, since Eclipse can't handle them properly (line spacing increases suddenly in those files). But I can't stop progress, I guess....

PR #904, opened 2015-02-26
Prevent race condition in CallByNameParamAtCreationPresenter

sschaef:
Fixes #1002388


dragos:
Spurious failure. LGTM

PR #905, opened 2015-03-08
Fix NPE in Scala Reconciler

sschaef:
`getDocumentProvider` returns `null` if the editor is closed.

Fixes #1002396

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/ui/editor/DecoratedInteractiveEditor.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/ui/editor/DecoratedInteractiveEditor.scala
@@ -15,21 +15,21 @@ import org.scalaide.util.ui.DisplayThread
 trait DecoratedInteractiveEditor extends ISourceViewerEditor {
 
   /** Return the annotation model associated with the current document. */
-  private def annotationModel = getDocumentProvider.getAnnotationModel(getEditorInput).asInstanceOf[IAnnotationModel]
+  private def annotationModel = Option(getDocumentProvider).map(_.getAnnotationModel(getEditorInput).asInstanceOf[IAnnotationModel])
 
   private var previousAnnotations = List[Annotation]()
 
   /**
    * Update annotations on the editor from a list of IProblems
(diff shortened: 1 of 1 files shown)


org.scala-ide.sdt.core/src/org/scalaide/ui/editor/DecoratedInteractiveEditor.scala
```diff
@@ -23,28 +23,30 @@ trait DecoratedInteractiveEditor extends ISourceViewerEditor {
...
-        // TODO: This should be replaced by a better modularization of semantic highlighting PositionsChange
-        val newPositions = newAnnotations.values
-        def end (x:Position) = x.offset + x.length - 1
-        val taintedBounds : (Int, Int) = ((Int.MaxValue, 0) /: newPositions) {(acc, p1) => (Math.min(acc._1, p1.offset), Math.max(acc._2, end(p1)))}
-        val taintedLength = (taintedBounds._2 - taintedBounds._1 +1)
-
-        DisplayThread.asyncExec { presViewer.asInstanceOf[ITextViewerExtension2].invalidateTextPresentation(taintedBounds._1, taintedLength) }
-    } else {
-        DisplayThread.asyncExec { getViewer.invalidateTextPresentation() }
+    annotationModel foreach { model ⇒
```
Kwestor: If you use this version:
`def updateErrorAnnotations(errors: List[IProblem], cu: ICompilationUnit): Unit = annotationModel foreach { model ⇒` all those lines will be left unchanged (which helps with reviews and file history browsing). But that's just a side-note.

Kwestor:
With small nit-pick, but LGTM.

org.scala-ide.sdt.core/src/org/scalaide/ui/editor/DecoratedInteractiveEditor.scala
sschaef: Yes, that is what I did now.

sschaef:
I should delete the branch after a merge and not before it...

PR #908, opened 2015-03-13
Handle single empty line in `RemoveDuplicatedEmptyLines` save action

sschaef:
Seq.sliding(n) returns a seq whose length can be smaller than n if less
than n elements exist.

Fixes #1002400

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/extensions/saveactions/RemoveDuplicatedEmptyLines.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/extensions/saveactions/RemoveDuplicatedEmptyLines.scala
@@ -26,7 +26,7 @@ trait RemoveDuplicatedEmptyLines extends SaveAction with DocumentSupport {
       case (r, _) => r.trimRight(document).length == 0
     }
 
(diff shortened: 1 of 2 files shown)


org.scala-ide.sdt.core/src/org/scalaide/extensions/saveactions/RemoveDuplicatedEmptyLines.scala
```diff
@@ -34,6 +34,9 @@ trait RemoveDuplicatedEmptyLines extends SaveAction with DocumentSupport {
           Seq()
     }
 
-    removedLines.toList
+    if (emptyLines.length == 1)
```
dragos: It's not obvious to me if `emptyLines` is a linear sequence (thus `length` can be slow), or not. If not, LGTM :)

sschaef: Indeed, I'm going to improve that.

sschaef:
spurious test failure

sschaef:
retest this please

PR #910, opened 2015-03-15
Fix warnings

sschaef:
Our number of warnings decreased from ~180 to ~110 (at least in my configuration). Half of the remaining warnings belong to deprecations in `sdt.core/plugin.xml`. The remaining ones contain deprecation warnings about the Eclipse API, all sort of warnings in the old wizard and in the indentation engine (which I both didn't fix because I want to kill both of them entirely) and scalac false positives.

This shouldn't introduce any behavior changes.


Kwestor:
LGTM, nice work, remove all those pesky buggers.

mpociecha:
LGTM

PR #912, opened 2015-03-16
Simplify/fix editor API.

dragos:
This commit should preserve binary compatibility so it can be included in 4.1.

It makes ICUs not assume the project presentation compiler, and this allows a potential Sbt editor to
use its own without needing to overwrite a lot of other methods.

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/core/completion/ScalaCompletions.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/core/completion/ScalaCompletions.scala
@@ -25,11 +25,21 @@ import org.scalaide.util.internal.Commons
 class ScalaCompletions extends HasLogger {
   import org.eclipse.jface.text.IRegion
 
-  /** Returns the list of possible completions, at the given position in the compilation unit,
-   *  with the given region as completion prefix.
-   */
+  @deprecated("Use getCompletions, which correctly handles units with a different presentation compiler", since = "4.1.0")
   def findCompletions(region: IRegion, position: Int, icu: InteractiveCompilationUnit): List[CompletionProposal] =
     icu.scalaProject.presentationCompiler { compiler =>
+      findCompletion(region, position, icu, compiler)
+    }.getOrElse(Nil)
+
+
+  /** Returns the list of possible completions, at the given position in the compilation unit,
+   *  with the given region as completion prefix.
+   *
+   *  @note This supports compilation units that have a different presentation compiler than the default
+   *        per-project compiler
+   */
+  def getCompletions(region: IRegion, position: Int, icu: InteractiveCompilationUnit): List[CompletionProposal] =
+    icu.presentationCompiler { compiler =>
       findCompletion(region, position, icu, compiler)
     }.getOrElse(Nil)
 
--- a/org.scala-ide.sdt.core/src/org/scalaide/ui/editor/DefaultScalaEditorConfiguration.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/ui/editor/DefaultScalaEditorConfiguration.scala
@@ -1,0 +1,104 @@
+package org.scalaide.ui.editor
+
+import org.eclipse.jface.text.source.IAnnotationHover
+import org.eclipse.jface.preference.IPreferenceStore
+import org.eclipse.jface.text.reconciler.IReconciler
+import org.eclipse.jface.text.reconciler.MonoReconciler
+import org.eclipse.jface.text.source.DefaultAnnotationHover
+import org.scalaide.core.lexical.ScalaPartitions
+import org.eclipse.jface.text.presentation.PresentationReconciler
+import org.scalaide.core.lexical.ScalaCodeScanners
+import org.eclipse.jface.text.source.SourceViewerConfiguration
+import org.eclipse.jface.text.information.InformationPresenter
+import org.eclipse.jdt.ui.text.IJavaPartitions
+import org.eclipse.jface.text.IDocumentListener
+import org.eclipse.jface.text.source.ISourceViewer
+import org.eclipse.jface.text.IDocument
+import org.scalaide.ui.editor.hover.IScalaHover
+import org.eclipse.jface.text.ITextHover
+import org.scalaide.core.IScalaPlugin
+import org.eclipse.jface.text.DocumentEvent
+import org.eclipse.jface.util.PropertyChangeEvent
+import org.eclipse.jface.text.rules.DefaultDamagerRepairer
+
+/** A default SourceViewerConfiguration class for Scala-based editors.
+ *
+ *  This class can be subclassed. It provides a default configuration including
+ *  reconciliation (errors-as-you-type), hovers and Scala syntax highlighting.
+ *
+ *  The workhorse of this implementation is [[InteractiveCompilationUnit]], which provides
+ *  the mapping between original source and Scala-source.
+ *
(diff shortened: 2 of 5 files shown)


org.scala-ide.sdt.core/src/org/scalaide/core/completion/ScalaCompletions.scala
```diff
@@ -25,11 +25,21 @@ import org.scalaide.util.internal.Commons
...
+
   /** Returns the list of possible completions, at the given position in the compilation unit,
    *  with the given region as completion prefix.
+   *
+   *  @note This supports compilation units that have a different presentation compiler than the default
+   *        per-project compiler
    */
-  def findCompletions(region: IRegion, position: Int, icu: InteractiveCompilationUnit): List[CompletionProposal] =
-    icu.scalaProject.presentationCompiler { compiler =>
+  def getCompletions(region: IRegion, position: Int, icu: InteractiveCompilationUnit): List[CompletionProposal] =
```
sschaef: Why not just update `findCompletions` with the new implementation?

org.scala-ide.sdt.core/src/org/scalaide/ui/editor/DefaultScalaEditorConfiguration.scala
```diff
@@ -0,0 +1,104 @@
...
+ *  the mapping between original source and Scala-source.
+ *
+ *  Any method can be overwritten.
+ *
+ *  @note Standard content assist is missing from this configuration, but should be easy to add
+ *        in a subsequent release
+ *
+ *  @since 4.1.0
+ */
+trait DefaultScalaEditorConfiguration extends SourceViewerConfiguration {
```
sschaef: I don't like this implementation right now. We already have 3 editor configurations right now (scala core, worksheet, play2), this would add a new one. We should use this one to unify all of them. Most stuff of the SSVC could be moved to this implementation, which then can be mixed into SSVC. This way we would also update the worksheet+play2 to the behavior of the SSVC.

org.scala-ide.sdt.core/src/org/scalaide/core/completion/ScalaCompletions.scala
dragos: Because of binary incompatibilities. The new method in `InteractiveCompilationUnit` is not going to be available on subclasses that compiled against the 4.0 version. So we can't add a call to the new method on an old code path, or clients will see `NoSuchMethodErrors`.

dragos:
This is an attempt to unify them. Right now, source configuration is duplicating a lot of code. So the first step is to have a default in place (this one). I agree that the `core` source configuration could mix this one here.

I could open PRs on worksheet and play2 to use this source configuration, but I don't think this is a blocker: we can make those changes at any point.

Let me know what you think

sschaef:
Generally, I would prefer if the new source configuration would implement more definitions from the main source configuration, but given that it can be easily improved in future, this LGTM.

org.scala-ide.sdt.core/src/org/scalaide/core/completion/ScalaCompletions.scala
sschaef: Ok. Though, I don't understand it - I guess one day I have to learn how binary compatibility in Scala really works.

dragos:
Good point, I'll see what else makes sense. I'd probably go for completions as well, but they need to handle source mappings correctly, so I'd need to add that too.

org.scala-ide.sdt.core/src/org/scalaide/core/completion/ScalaCompletions.scala
dragos: The problem is the JVM: a trait becomes an interface and an implementation class. In this case:

```scala
interface IncrementalCompilationUnit {
..
  def presentationCompiler: IScalaPresentationCompilerProxy
}

class  IncrementalCompilationUnit$class {
  static def presentationCompiler($this: IncrementalCompilationUnit): IScalaPresentationCompilerProxy = // my implementation
}
```

every subclass of this trait will get a synthetic implementation for the interface method. For example, for play2:

```scala
class Play2CompilationUnit extends ... {
  // forwarder
  def presentationCompiler = IncrementalCompilationUnit$class.presentationCompiler(this)
}
```

So, changing the interface makes all subclasses inherit a new (abstract) method, and the forward will be missing. That's ok as long as nobody is calling that method, but in this PR I also add a *call* to the new method. So, unless subclasses recompile and republish, they'd be broken.

sschaef: Ok, thanks for the explanation.

org.scala-ide.sdt.core/src/org/scalaide/ui/editor/DefaultScalaEditorConfiguration.scala
```diff
@@ -0,0 +1,104 @@
...
+import org.eclipse.jdt.ui.text.IJavaPartitions
+import org.eclipse.jface.text.IDocumentListener
+import org.eclipse.jface.text.source.ISourceViewer
+import org.eclipse.jface.text.IDocument
+import org.scalaide.ui.editor.hover.IScalaHover
+import org.eclipse.jface.text.ITextHover
+import org.scalaide.core.IScalaPlugin
+import org.eclipse.jface.text.DocumentEvent
+import org.eclipse.jface.util.PropertyChangeEvent
+import org.eclipse.jface.text.rules.DefaultDamagerRepairer
```
Kwestor: What is the official standpoint for imports in codebase? Should we organize them all?

org.scala-ide.sdt.core/src/org/scalaide/ui/editor/ReconcilingStrategy.scala
```diff
@@ -25,7 +30,7 @@ class ReconcilingStrategy(sourceEditor: SourceCodeEditor, documentListener: IDoc
   override def reconcile(partition: IRegion) {
     for (doc <- document) {
       val errors = sourceEditor.getInteractiveCompilationUnit.forceReconcile()
-      sourceEditor.updateErrorAnnotations(errors)
+      sourceEditor.updateErrorAnnotations(errors, null)
```
Kwestor: Why `null`? Maybe name this parameter?

org.scala-ide.sdt.core/src/org/scalaide/ui/editor/DefaultScalaEditorConfiguration.scala
sschaef: I'm for organizing them, but as long as organize imports is buggy, we can't force/half-force it with a save action.

dragos: I'm also in favour of using organise imports (I'll do that here).

org.scala-ide.sdt.core/src/org/scalaide/ui/editor/ReconcilingStrategy.scala
dragos: Good point. I'l name it. It's the JDT JavaElement corresponding to this source, and I forgot exactly when it's used. In this case, there is no such element, but the API expects it anyway.

dragos:
retest this please

sschaef:
Superseded by #1095.

PR #914, opened 2015-03-19
Mark some tests as flaky

sschaef:
The tests failed lately on the build server.

Also: Rename 'witn' to 'with'


Kwestor:
I remember those failing for me also.

Kwestor:
LGTM

mpociecha:
Out of curiosity - does it mean that these features 'marked' as flaky can randomly fail also when using them in installed Scala IDE?

sschaef:
I'm not entirely sure about that. I think it is only a problem in our test suite because the compiler gets confused by the auto generated documents (which sometimes share the same package/type names and for speedup reasons we don't restart the compiler after every test). The only feature I know of that can randomly fail in production are refactorings but I don't know why this happens.

PR #917, opened 2015-04-01
old version number and transitive deps

skyluc:
Updates the build with the latest Scala versions.
Removes the transitive dependencies to the dependencies of the Scala 2.10 fat jar. Tycho is making a mess with these when trying to resolve the p2 dependencies, and they are not useful.

### DIFF
--- a/org.scala-ide.scala210.jars/pom.xml
+++ b/org.scala-ide.scala210.jars/pom.xml
@@ -17,28 +17,33 @@
       <groupId>org.scala-lang</groupId>
       <artifactId>scala-library</artifactId>
       <version>${scala210.version}</version>
+      <optional>true</optional>
     </dependency>
     <dependency>
       <groupId>org.scala-lang</groupId>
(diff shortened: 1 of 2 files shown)


dragos:
I think this is a good change, but what I don't understand was how come this dependency was still pulled.... because the uber-build sets the `SCALA210_VERSION` to `2.10.4`, so this variable should have been overridden, right?

skyluc:
The answer to @dragos question: it looks like Tycho is making a mess while making its dependencies resolution. The pom.xml files are read, but the profiles are not applied, so the default values are used, and for the Scala version, the default value was 2.10.5-SNAPSHOT.

dragos:
@lrytz let us know if this helped.

lrytz:
It helped, thank you @skyluc and @dragos! Recent builds don't mention 2.10.5-SNAPSHOT anymore (https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-ide/715/consoleFull)

PR #918, opened 2015-04-01
Make organize imports format consistent with scalariform

misto:
We used to pass the IDE's SpaceInsideParentheses setting to scala-refactoring to control the spaces inside the braces of multiple imports, like this:

import a.b.{ c, d }

Unfortunately, in scalariform SpaceInsideParentheses only affects the space around regular parenthesis, and there is no setting for the spaces in imports. So to make the code generated by scala-refactoring consistent with the formatting of scalariform, we always print a space inside multiple imports.

Fixes scala-ide/scala-refactoring#81.


sschaef:
It seems the bot is in a hurry.

sschaef:
LGTM

PR #919, opened 2015-04-06
Also apply refactoring to files not currently open in the editor

mlangc:
`TextFileChange.LEAVE_DIRTY` causes changes to files that are not currently open
in the editor to be silently ignored. `TextFileChange.FORCE_SAVE` seems more appropriate if the flag `saveAfter` is set anyway.

Fixes #1002079

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/refactoring/internal/OrganizeImports.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/refactoring/internal/OrganizeImports.scala
@@ -218,6 +218,8 @@ class OrganizeImports extends RefactoringExecutorWithoutWizard {
 
     val refactoring = withCompiler( c => new implementations.OrganizeImports with FormattingOverrides { val global = c })
 
+    override protected def leaveDirty = true
+
     override def checkInitialConditions(pm: IProgressMonitor) = {
       val status = super.checkInitialConditions(pm)
       if(compilationUnitHasProblems) {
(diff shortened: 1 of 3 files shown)


sschaef:
It is interesting that this change fixes the problem. Unfortunately the fix is not so easy, the `saveAfter` was introduced by purpose: https://github.com/scala-ide/scala-ide/commit/44fbc5f73d

sschaef:
In the original discussion I even saw that problem: https://github.com/scala-ide/scala-ide/pull/653

sschaef:
It seems that the fix should be
```
setSaveMode(if (saveAfter) TextFileChange.FORCE_SAVE else TextFileChange.LEAVE_DIRTY)
```
But organize imports sets the `saveAfter` flag to true, which shouldn't be the case.

mlangc:
Yea, I feared that it couldn't be all that easy... here is a second try ;-).

sschaef:
By accident, you didn't find out why the refactoring isn't written to disk when the editor is left dirty? I guess the refactorings are performed, but only to the working copies, which we then might discard somehow.

sschaef:
Also, try to not reference a non merged commit hash in a commit message. If someone is doing a rebase, the hash is no longer valid.

mlangc:
As for the commit: Yea, I know - in that case I've to amend the commit... but I could also write 'the previous commit' - or merge them altogether.

As for finding why leaving files dirty causes changes to be lost: I guess it's in parts a conceptual problem. How should a file that is not opened in an editor be left dirty? In my opinion there are only two solutions:

1. Open an editor with the changed file that is marked as "dirty".
2. Silently discard the changes.

It seems that Eclipse took option 2.

mlangc:
As for Eclipse taking option 2: I think the relevant code that confirms this can be found here: `org.eclipse.ltk.core.refactoring.TextFileChange.needsSaving()`

sschaef:
Ok, thanks. That makes sense.

sschaef:
LGTM

sschaef:
Just in case you are interested to continue working on this, here is a follow-up ticket: https://www.assembla.com/spaces/scala-ide/tickets/1002427-restart-presentation-compiler-after-performed-rename-refactoring#/activity/ticket:

PR #921, opened 2015-04-09
Basic support for Hot Code Replacement in debugged JVM

mpociecha:
This PR adds the possibility to modify and compile code, when an application is run in a debug mode, and to have these changes visible and taken into account by the debugged VM without restarting the application. This feature can be turned on/off (right now it's turned off by default). User can decide whether files containing compilation errors should be skipped. Stack frames affected by changes can be automatically dropped. Sometimes users don't like automatic dropping frames after HCR and they prefer to do that manually (see HCR in IntelliJ). Right now there's added also such a possibility. On the other hand it's a question whether it should be available, if we'll know it can cause problems. Therefore, it may be removed in the future, if the HCR with automatic dropping frames would be much more reliable. There are also added several integration JUnit tests.

Note that HCR is limited by JVM's support for replacing classes. Its goal is to allow to experiment with the code inside methods/blocks rather than to develop the whole application in the interactive mode. Changes of methods' signatures etc. don't work - similarly to other HCR implementations from Eclipse or other IDEs.

The implementation is inspired by the HCR support for Java. Scala debug target, threads etc. don't extend the default Java/JDI model and build their own one. Therefore, they are skipped when processing events by various Java-focused managers.

There's added ScalaHotCodeReplaceManager which implements HCR support using the Scala model. Every ScalaDebugTarget has its instance of manager which is also the listener waiting for events related to changed resources. When we get an event about changed class files, we get byte arrays containing new versions of these files and replace the old ones in VM. After this operation it's needed to rebind frames (optionally also drop the obsolete ones) and to refresh breakpoints.

Each from several added tests contains a fragment where the code is modified and recompiled. To ensure that all events and messages are correctly propagated after the additional incremental build, it was needed to use Thread.sleep with 500 millis. I know it's quite ugly solution but it seems to work well and there are not many such tests.

There are serious reasons why this implementation is called the basic support:

- Automatic dropping frames is based on the information which frames are marked as obsolete.

- We get this information after performing HCR but it can cause problems. According to the comment from Eclipse's HCR implementation for Java, JVM is faulty when there are frames rendered as obsolete. Therefore, in Java implementation they try to find affected frames and drop them before performing HCR. We should do the same thing because VM sometimes crashes after HCR and some additional operations. This is the serious blocker. Anyway even current implementation can be quite helpful in many situations. Except JUnit tests it was used to debug e.g.some multithreaded application using futures.

- There's also another problem. After performing HCR sometimes already existing breakpoints behave strange. I had situations where from time to time I was getting a JDI event for the line breakpoint with the different location than the line specified for this breakpoint (sometimes it was one line before or after the expected one). I was investigating this for a while but didn't figure out what to do with that. Also I wasn't able to reproduce such a behaviour in tests. It's something, what should be checked once again after the addition of the new approach to drop affected frames. In case of problems it's always possible to recreate given frame manually, drop to the above frame and go to the expected breakpoint once again. Note: it's sometimes hard to hit breakpoints in Scala IDE even without HCR (they quite often are nondeterministic from the end-user's point of view of). Hence, it's not so simple to test this. It's not possible to fix everything, which basically is not related to this particular feature, but is also used when debugging.

- Dialogs shown when replacing classes is not supported by VM or fails don't contain buttons to terminate or restart debug session. Currently there's only an adequate message.

Lastly, the below disclaimer is required by the lawyers:

THE FOLLOWING DISCLAIMER APPLIES TO ALL SOFTWARE CODE AND OTHER MATERIALS CONTRIBUTED IN CONNECTION WITH THIS SOFTWARE:
THIS SOFTWARE IS LICENSED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS “AS IS” AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE AND ANY WARRANTY OF NON-INFRINGEMENT, ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE. THIS SOFTWARE MAY BE REDISTRIBUTED TO OTHERS ONLY BY EFFECTIVELY USING THIS OR ANOTHER EQUIVALENT DISCLAIMER IN ADDITION TO ANY OTHER REQUIRED LICENSE TERMS.
ONLY THE SOFTWARE CODE AND OTHER MATERIALS CONTRIBUTED IN CONNECTION WITH THIS SOFTWARE, IF ANY, THAT ARE ATTACHED TO (OR OTHERWISE ACCOMPANY) THIS SUBMISSION (AND ORDINARY COURSE CONTRIBUTIONS OF FUTURES PATCHES THERETO) ARE TO BE CONSIDERED A CONTRIBUTION. NO OTHER SOFTWARE CODE OR MATERIALS ARE A CONTRIBUTION.

### DIFF
--- a/org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/breakpoints/ScalaDebugBreakpointManager.scala
+++ b/org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/breakpoints/ScalaDebugBreakpointManager.scala
@@ -5,6 +5,7 @@ import org.eclipse.core.resources.IMarkerDelta
 import org.eclipse.debug.core.DebugPlugin
 import org.eclipse.debug.core.IBreakpointListener
 import org.eclipse.debug.core.model.IBreakpoint
+import org.eclipse.jdt.internal.debug.core.breakpoints.JavaLineBreakpoint
 import scala.actors.Actor
 import BreakpointSupportActor.Changed
 import org.scalaide.debug.internal.BaseDebuggerActor
@@ -50,6 +51,13 @@ class ScalaDebugBreakpointManager private (/*public field only for testing purpo
     companionActor ! BreakpointAdded(breakpoint)
   }
 
+  /**
+   * Intended to ensure that we'll hit already defined and enabled breakpoints after performing hcr.
+   * @param changedClassesNames fully qualified names of types
+   */
+  def reenableBreakpointsInClasses(changedClassesNames: Seq[String]): Unit =
+    companionActor ! ReenableBreakpointsAfterHcr(changedClassesNames)
+
   // ------------
 
   def init() {
@@ -88,6 +96,8 @@ private[debug] object ScalaDebugBreakpointManagerActor {
   case class BreakpointAdded(breakpoint: IBreakpoint)
   case class BreakpointRemoved(breakpoint: IBreakpoint)
   case class BreakpointChanged(breakpoint: IBreakpoint, delta: IMarkerDelta)
+  case class ReenableBreakpointsAfterHcr(classNames: Seq[String])
+  case object ReenableBreakpointAfterHcr
 
   private final val JdtDebugUID = "org.eclipse.jdt.debug"
 
@@ -138,6 +148,8 @@ private class ScalaDebugBreakpointManagerActor private(debugTarget: ScalaDebugTa
         case _ =>
           // see previous comment
       }
+    case ReenableBreakpointsAfterHcr(changedClassesNames) =>
+      reenableBreakpointAfterHcr(changedClassesNames)
     case ScalaDebugBreakpointManager.ActorDebug =>
       reply(None)
 
@@ -151,6 +163,33 @@ private class ScalaDebugBreakpointManagerActor private(debugTarget: ScalaDebugTa
 
   private def createBreakpointSupport(breakpoint: IBreakpoint): Unit = {
     breakpoints += (breakpoint -> BreakpointSupport(breakpoint, debugTarget))
+  }
+
+  private def reenableBreakpointAfterHcr(changedClassesNames: Seq[String]): Unit = {
+    /*
+     * We need to prepare names of changed classes and these taken from breakpoints because
+     * from some reasons they differ. We need to change them slightly as:
+     *
+     * Type names used in breakpoints from some reason have double intermediate dollars,
+     * e.g. debug.Foo$$x$$Bar instead of debug.Foo$x$Bar, debug.Foo$$x$ instead of debug.Foo$x$.
+     *
+     * There are also anonymous types which really should have double dollars but anyway
+     * breakpoints for such types have currently set type like
+     * com.test.debug.Foo$$x$$Bar$java.lang.Object$java.lang.Object
+     * instead of
+     * debug.Foo$x$Bar$$anon$2$$anon$1
+     */
+    val anonTypePattern = """\$anon\$[1-9][0-9]*"""
+    val namesToCompareWithOnesFromBreakpoints = changedClassesNames.map(_.replaceAll(anonTypePattern, "java.lang.Object"))
+    def isChanged(typeName: String): Boolean =
+      namesToCompareWithOnesFromBreakpoints.contains(typeName.replace("$$", "$"))
+
+    val affectedBreakpoints = breakpoints.keys.collect {
+      case bp: JavaLineBreakpoint if isChanged(bp.getTypeName) => bp
+    }
+    affectedBreakpoints.foreach { breakpoint =>
+      breakpoints(breakpoint) ! ReenableBreakpointAfterHcr
+    }
   }
 
   override protected def preExit(): Unit = breakpoints.values.foreach(_ ! PoisonPill)
--- a/org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/breakpoints/ScalaDebugBreakpointManager.scala
+++ b/org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/breakpoints/ScalaDebugBreakpointManager.scala
@@ -7,7 +7,6 @@ import org.eclipse.debug.core.IBreakpointListener
 import org.eclipse.debug.core.model.IBreakpoint
 import org.eclipse.jdt.internal.debug.core.breakpoints.JavaLineBreakpoint
 import scala.actors.Actor
-import BreakpointSupportActor.Changed
 import org.scalaide.debug.internal.BaseDebuggerActor
 import org.scalaide.debug.internal.PoisonPill
 
@@ -96,8 +95,9 @@ private[debug] object ScalaDebugBreakpointManagerActor {
   case class BreakpointAdded(breakpoint: IBreakpoint)
   case class BreakpointRemoved(breakpoint: IBreakpoint)
   case class BreakpointChanged(breakpoint: IBreakpoint, delta: IMarkerDelta)
+
+  /** The message used to reenable all breakpoints related to given classes. */
   case class ReenableBreakpointsAfterHcr(classNames: Seq[String])
-  case object ReenableBreakpointAfterHcr
 
   private final val JdtDebugUID = "org.eclipse.jdt.debug"
 
@@ -111,6 +111,7 @@ private[debug] object ScalaDebugBreakpointManagerActor {
 private class ScalaDebugBreakpointManagerActor private(debugTarget: ScalaDebugTarget) extends BaseDebuggerActor {
   import ScalaDebugBreakpointManagerActor._
   import BreakpointSupportActor.Changed
+  import BreakpointSupportActor.ReenableBreakpointAfterHcr
 
   private var breakpoints = Map[IBreakpoint, Actor]()
 
@@ -168,9 +169,9 @@ private class ScalaDebugBreakpointManagerActor private(debugTarget: ScalaDebugTa
   private def reenableBreakpointAfterHcr(changedClassesNames: Seq[String]): Unit = {
     /*
      * We need to prepare names of changed classes and these taken from breakpoints because
-     * from some reasons they differ. We need to change them slightly as:
+     * for some reasons they differ. We need to change them slightly as:
      *
-     * Type names used in breakpoints from some reason have double intermediate dollars,
+     * Type names used in breakpoints have double intermediate dollars,
      * e.g. debug.Foo$$x$$Bar instead of debug.Foo$x$Bar, debug.Foo$$x$ instead of debug.Foo$x$.
      *
      * There are also anonymous types which really should have double dollars but anyway
--- a/org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/hcr/ChangedClassFilesVisitor.scala
+++ b/org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/hcr/ChangedClassFilesVisitor.scala
@@ -17,8 +17,9 @@ import org.eclipse.jdt.core.JavaCore
 import org.eclipse.jdt.core.ToolFactory
 import org.eclipse.jdt.core.util.IClassFileReader
 import org.eclipse.jdt.internal.debug.core.JavaDebugUtils
-import org.scalaide.logging.HasLogger
 import org.scalaide.debug.internal.preferences.HotCodeReplacePreferences
+import org.scalaide.logging.HasLogger
+import org.scalaide.util.eclipse.EclipseUtils
 
 private[internal] case class ClassFileResource(fullyQualifiedName: String, classFile: IFile)
 
@@ -34,7 +35,7 @@ private[internal] class ChangedClassFilesVisitor extends IResourceDeltaVisitor w
 
   // the value stored as a field to finish applying the same strategy to all classes even when someone changed
   // the configuration in the meantime
-  private var replaceDespiteCompilationErrors = HotCodeReplacePreferences.performHcrForFilesContainingErrors
+  private val replaceDespiteCompilationErrors = HotCodeReplacePreferences.performHcrForFilesContainingErrors
 
   /**
    * Found classes containing changes.
@@ -42,14 +43,6 @@ private[internal] class ChangedClassFilesVisitor extends IResourceDeltaVisitor w
   private val changedClasses = scala.collection.mutable.Set[ClassFileResource]()
 
   def getChangedClasses: List[ClassFileResource] = changedClasses.toList
-
-  /**
-   * It should be called always before the visitor is reused.
-   */
-  def reset(): Unit = {
-    replaceDespiteCompilationErrors = HotCodeReplacePreferences.performHcrForFilesContainingErrors
-    changedClasses.clear()
-  }
 
   /**
    * Looks for modified classes, adds them to changedClasses and decides whether children should be visited.
@@ -76,6 +69,8 @@ private[internal] class ChangedClassFilesVisitor extends IResourceDeltaVisitor w
     } {
       val slashDelimitedQualifiedName = new String(reader.getClassName())
 
+      // Replacing classes, when there are errors, can lead to the broken debug session (like in Java).
+      // We have it for consistency with Java HCR implementation.
       if (replaceDespiteCompilationErrors || !hasCompilationErrors(classFile, reader, slashDelimitedQualifiedName)) {
         val className = slashDelimitedQualifiedName.replace('/', '.')
         changedClasses.add(ClassFileResource(className, classFile))
@@ -95,7 +90,7 @@ private[internal] class ChangedClassFilesVisitor extends IResourceDeltaVisitor w
    * @param fullyQualifiedName slash delimited name of type
    */
   private def hasCompilationErrors(classFile: IFile, reader: IClassFileReader, fullyQualifiedName: String): Boolean =
-    try {
+    EclipseUtils.withSafeRunner(s"Error occurred while looking for compilation error markers for type '$fullyQualifiedName'") {
       def isErrorMarker(marker: IMarker): Boolean =
         marker.getAttribute(IMarker.SEVERITY, /* defaultValue = */ IMarker.SEVERITY_INFO) == IMarker.SEVERITY_ERROR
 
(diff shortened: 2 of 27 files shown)


mpociecha:
@dragos @sschaef Could you take a look?

misto:
ok to test

sschaef:
add to whitelist

sschaef:
After EE is merged, this contains some merge errors.

mpociecha:
I'll rebase it tomorrow.

mpociecha:
I assume someone will look at Jenkins (to check what causes slowdowns) or will increase time limits. Lately all PRs were failing because of that.

sschaef:
@dragos already fixed it

sschaef:
retest this please

mpociecha:
The build for Luna succeeded (Total time: 29:09.817s) and the build for Kepler has been aborted at its very end (Build timed out after 30 minutes).

dragos:
retest this pleaase

dragos:
I just increased the timeout to be 3x the average successful build. If all fails, I'll set an absolute timeout value. It seems the eval expression tests have bumped the build time well above the previous times.

mpociecha:
The build didn't start. Let me check whether I can run it by myself.

mpociecha:
retest this please

sschaef:
I just reviewed the functionality.

The only ting I have to say about this is:

awesome

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/breakpoints/ScalaDebugBreakpointManager.scala
```diff
@@ -88,6 +96,8 @@ private[debug] object ScalaDebugBreakpointManagerActor {
   case class BreakpointAdded(breakpoint: IBreakpoint)
   case class BreakpointRemoved(breakpoint: IBreakpoint)
   case class BreakpointChanged(breakpoint: IBreakpoint, delta: IMarkerDelta)
+  case class ReenableBreakpointsAfterHcr(classNames: Seq[String])
+  case object ReenableBreakpointAfterHcr
```
dragos: Is this case object used? Even if it is used, can't the same semantics be attached to the case class, when `classNames` is empty?

dragos: Or, if both are needed, maybe add a short scaladoc to describe when each one should be used.

mpociecha:
In the future I'm going to add the semantic dropping frames before HCR (to avoid VM crashes). Right now I'm not sure how to process deltas and how to extract helpful information, but I will investigate this.

sschaef:
Maybe this question gets solved when I look in more detail to the code, but why is it possible right now to replace a nested def (def inside def) but not a private def (def inside class)?

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/breakpoints/ScalaDebugBreakpointManager.scala
mpociecha: The class is used when sending a message to `ScalaDebugBreakpointManagerActor` and the object is used when sending a message to `BreakpointSupportActor` from `ScalaDebugBreakpointManagerActor` (in this case we don't need to add an additional information).

In other words, it's not that the object is used, when classNames is empty. When it's empty, we don't need to perform HCR. Hence, we also don't need to send these messages.

Do you prefer to describe them or maybe simply the object should be moved to `BreakpointSupportActor` object? WDYT?

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/hcr/ChangedClassFilesVisitor.scala
```diff
@@ -0,0 +1,156 @@
...
+   * Found classes containing changes.
+   */
+  private val changedClasses = scala.collection.mutable.Set[ClassFileResource]()
+
+  def getChangedClasses: List[ClassFileResource] = changedClasses.toList
+
+  /**
+   * It should be called always before the visitor is reused.
+   */
+  def reset(): Unit = {
```
dragos: I don't really like the mutability in this class. I find it simpler to make the class "one-time-use" than to remember to call reset from time to time. Clients simply instantiate another one (there doesn't seem to be any expensive operation in the constructor, so the cost is negligible).

mpociecha:
@sschaef Do you mean the body or the whole def (its signature and so on)?
Def inside class works, when you change only the body. At least in my test cases.

If you change also the signature, probably all cases will fail. There's a dialog with an error message shown.

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/model/ScalaDebugTarget.scala
```diff
@@ -158,8 +166,14 @@ abstract class ScalaDebugTarget private(val virtualMachine: VirtualMachine,
   @volatile
   private var threads = List[ScalaThread]()
 
+  @volatile
+  private var isDuringHcr: Boolean = false
+  private[internal] def isPerformingHotCodeReplace: Boolean = isDuringHcr
+  private[internal] def isPerformingHotCodeReplace_=(isPerforming: Boolean) = { isDuringHcr = isPerforming }
```
dragos: I'm not sure what is the difference between having a `var` and having a getter *and* a setter for a private var. In this case it seems there is none.

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/model/ScalaThread.scala
```diff
@@ -47,10 +48,12 @@ abstract class ScalaThread private(target: ScalaDebugTarget, val threadRef: Thre
...
   // Members declared in org.eclipse.debug.core.model.IStep
 
-  override def canStepInto: Boolean = suspended // TODO: need real logic
-  override def canStepOver: Boolean = suspended // TODO: need real logic
-  override def canStepReturn: Boolean = suspended // TODO: need real logic
+  override def canStepInto: Boolean = canStep
+  override def canStepOver: Boolean = canStep
+  override def canStepReturn: Boolean = canStep
   override def isStepping: Boolean = ???
+  private def canStep = suspended && // TODO: needs real logic
```
dragos: Is the `TODO` sill needed?

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/hcr/ChangedClassFilesVisitor.scala
mpociecha: Sure, it is negligible. I'll remove `reset` method and will change var to val.

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/model/ScalaThread.scala
```diff
@@ -62,7 +65,8 @@ abstract class ScalaThread private(target: ScalaDebugTarget, val threadRef: Thre
 
   // Members declared in org.eclipse.debug.core.model.ISuspendResume
 
-  override def canResume: Boolean = suspended // TODO: need real logic
+  override def canResume: Boolean = suspended && // TODO: needs real logic
```
dragos: Is the `TODO` sill needed?

sschaef:
On 04/14/2015 12:31 PM, Michał Pociecha wrote:
>
> @sschaef <https://github.com/sschaef> Do you mean the body or the 
> whole def (its signature and so on)?
>
I meant the signature. I can change the signature as I will as long as 
the def is nested.

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/breakpoints/ScalaDebugBreakpointManager.scala
```diff
@@ -153,5 +165,32 @@ private class ScalaDebugBreakpointManagerActor private(debugTarget: ScalaDebugTa
     breakpoints += (breakpoint -> BreakpointSupport(breakpoint, debugTarget))
   }
 
+  private def reenableBreakpointAfterHcr(changedClassesNames: Seq[String]): Unit = {
+    /*
+     * We need to prepare names of changed classes and these taken from breakpoints because
+     * from some reasons they differ. We need to change them slightly as:
```
sschaef: s/from/for/

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/breakpoints/ScalaDebugBreakpointManager.scala
dragos: Both, please :) Since actors are untyped, we need to use conventions, and I think it makes is much easier to group messages in the companion object of the actor that handles them. A one-line description is also useful, but not always necessary. I leave it up to do you decide if it's needed.

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/model/ScalaThread.scala
mpociecha: I don't know what is the real logic which an author had in  mind. I only added one condition. Thus I didn't remove TODOs.

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/breakpoints/ScalaDebugBreakpointManager.scala
```diff
@@ -153,5 +165,32 @@ private class ScalaDebugBreakpointManagerActor private(debugTarget: ScalaDebugTa
...
+     *
+     * Type names used in breakpoints from some reason have double intermediate dollars,
```
sschaef: s/from/for/

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/model/ScalaThread.scala
mpociecha: I don't know what is the real logic which an author had in mind. I only added one condition. Thus I didn't remove TODOs.

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/model/ScalaThread.scala
mpociecha: Same here.

dragos:
A few minor comments, otherwise I'm as excited as @sschaef!

One comment regarding spurious errors. These could be related to our thread-unsafe use of the JDI calls. As described in [1002160](https://www.assembla.com/spaces/scala-ide/tickets/1002160), which should eventually be fixed and make the debugger *way more* reliable.

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/hcr/ScalaHotCodeReplaceManager.scala
```diff
@@ -0,0 +1,167 @@
...
+    if (delta == null || event.getType() != IResourceChangeEvent.POST_BUILD)
+      Nil
+    else
+      try {
+        visitor.reset()
+        delta.accept(visitor)
+        visitor.getChangedClasses
+      } catch {
+        case e: Exception =>
+          logger.error(s"Something went wrong when processing an event with delta $delta", e)
```
sschaef: Does `delta` have a useful `toString` implementation?

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/hcr/ScalaHotCodeReplaceManager.scala
```diff
@@ -0,0 +1,167 @@
...
+    if (changedClasses.nonEmpty)
+      debugTargetCompanionActor ! ReplaceClasses(changedClasses)
+  }
+
+  private def getChangedClasses(event: IResourceChangeEvent): List[ClassFileResource] = {
+    val delta = event.getDelta()
+    if (delta == null || event.getType() != IResourceChangeEvent.POST_BUILD)
+      Nil
+    else
+      try {
```
sschaef: Instead of the `try` block, you could use `EclipseUtils.withSafeRunner`.

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/model/ScalaThread.scala
dragos: I think we should remove them. I can't think of a more elaborate logic right now :)

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/preferences/DebuggerPreferencePage.scala
```diff
@@ -59,6 +77,16 @@ object DebuggerPreferencePage {
   val FILTER_SETTER = BASE_FILTER + Setter
   val FILTER_DEFAULT_GETTER = BASE_FILTER + DefaultGetter
   val FILTER_FORWARDER = BASE_FILTER + Forwarder
+
+  val HotCodeReplaceEnabled = "org.scala-ide.sdt.debug.hcr.enabled"
+  val NotifyAboutFailedHcr = "org.scala-ide.sdt.debug.hcr.notifyFailed"
+  val NotifyAboutUnsupportedHcr = "org.scala-ide.sdt.debug.hcr.notifyUnsupported"
+
+  // like in Java, it can lead to broken debug session but we have it for consistency
+  // with Java's hcr implementation
```
sschaef: This comment should be placed at the logic that replaces code with errors, not here in the preference page.

mpociecha: Ok.

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/hcr/ScalaHotCodeReplaceManager.scala
mpociecha: It seems to me it was something like file path. But I'll check it once again.

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/breakpoints/ScalaDebugBreakpointManager.scala
mpociecha: I'll do both then. :)

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/model/ScalaThread.scala
mpociecha: Ok.

mpociecha:
Would you like to have corrections in the separate commit (for easier reviewing) or can I add them to the second commit?

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/model/ScalaDebugTarget.scala
mpociecha: It seems to me I was using different access modifiers for a while. Currently it doesn't make sense to have a getter and a setter. Thanks.

sschaef:
Create a new commit.

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/hcr/ScalaHotCodeReplaceManager.scala
mpociecha: My fault. It's always `/`. We can get affected projects using getAffectedChildren or probably even traverse the tree further.

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/hcr/ScalaHotCodeReplaceManager.scala
mpociecha: You're right, but then I would log errors using the Eclipse log - not the Scala IDE log.

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/hcr/ScalaHotCodeReplaceManager.scala
sschaef: I think we always want to report errors to users. They should report them, in order that we can fix them.

mpociecha: I guess it's rather an answer to "You're right, but then I would log errors using the Eclipse log - not the Scala IDE log.", right?

Do you still suggest to change this one to EclipseUtils.withSafeRunner? I can also change logger to eclipseLog in other places, where we catch errors, but wouldn't it be strange when we'll log some things to the Scala IDE log and errors to the Eclipse log?

sschaef: On 04/14/2015 07:36 PM, Michał Pociecha wrote:
>
> In 
> org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/hcr/ScalaHotCodeReplaceManager.scala 
> <https://github.com/scala-ide/scala-ide/pull/921#discussion_r28350018>:
>
> > +    if (changedClasses.nonEmpty)
> > +      debugTargetCompanionActor ! ReplaceClasses(changedClasses)
> > +  }
> > +
> > +  private def getChangedClasses(event: IResourceChangeEvent): List[ClassFileResource] = {
> > +    val delta = event.getDelta()
> > +    if (delta == null || event.getType() != IResourceChangeEvent.POST_BUILD)
> > +      Nil
> > +    else
> > +      try {
> > +        visitor.reset()
> > +        delta.accept(visitor)
> > +        visitor.getChangedClasses
> > +      } catch {
> > +        case e: Exception =>
> > +          logger.error(s"Something went wrong when processing an event with delta $delta", e)
>
> I guess it's rather an answer to "You're right, but then I would log 
> errors using the Eclipse log - not the Scala IDE log.", right?
>
> Do you still suggest to change this one to 
> EclipseUtils.withSafeRunner? I can also change logger to eclipseLog in 
> other places, where we catch errors, but wouldn't it be strange when 
> we'll log some things to the Scala IDE log and errors to the Eclipse log?
>
Changing it to `eclipseLog` is ok, there is another try-catch which 
can't be converted anyway. Actually, both `logger` and `eclipseLog` do 
log to the same underlying logger. But Everything that is reported to 
`eclipseLog` also gets displayed in the error view (and users may get 
interrupted by them, that depends on what they have configured).

mpociecha:
@dragos @sschaef That's it. I've already taken into account all your remarks.

sschaef:
Great, LGTM.

PR #922, opened 2015-04-12
Small Scaladoc corrections

sschaef:


mpociecha:
LGTM (regardless of current problems with Jenkins)

PR #923, opened 2015-04-12
Add Expression Evaluator to update site

sschaef:
I forgot to make the expression evaluator available to the update site
by adding it as feature entries to the corresponding configuration
files. Without these changes the EE is no part of the Scala IDE
installation.


mpociecha:
LGTM and it even works (verified)

PR #924, opened 2015-04-13
Make interpreter background color configurable

sschaef:
First commit is bugfix, second commit small improvements, third commit are cleanups. See commit messages for further information.

### DIFF
--- a/org.scala-ide.sdt.core/plugin.xml
+++ b/org.scala-ide.sdt.core/plugin.xml
@@ -1838,6 +1838,33 @@
           The font used for all Scala hovers.
        </description>
     </fontDefinition>
+    <colorDefinition
+          categoryId="org.scalaide.ui.theme"
+          id="org.scalaide.ui.color.interpreterForeground"
+          label="Interpreter foreground color"
+          value="60,0,128">
+       <description>
+          The color used for the Scala Interpreter foreground.
+       </description>
+    </colorDefinition>
+    <colorDefinition
+          categoryId="org.scalaide.ui.theme"
+          id="org.scalaide.ui.color.interpreterBackground"
+          label="Interpreter background color"
+          value="230,230,230">
+       <description>
+          The color used for the Scala Interpreter background.
+       </description>
+    </colorDefinition>
+    <colorDefinition
+          categoryId="org.scalaide.ui.theme"
+          id="org.scalaide.ui.color.interpreterErrorForeground"
+          label="Interpreter error foreground color"
+          value="128,0,64">
+       <description>
+          The color used for the Scala Interpreter foreground to display errors.
+       </description>
+    </colorDefinition>
  </extension>
  <extension
        point="org.scala-ide.sdt.core.semanticHighlightingParticipants">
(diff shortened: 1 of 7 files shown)


dragos:
retest this please

org.scala-ide.sdt.core/plugin.xml
```diff
@@ -1838,6 +1838,42 @@
           The font used for all Scala hovers.
        </description>
     </fontDefinition>
+    <colorDefinition
+          categoryId="org.scalaide.ui.theme"
+          id="org.scalaide.ui.color.interpreterForeground"
+          label="Interpreter foreground color"
+          value="60,0,128">
```
dragos: I think we need to use `defaultsTo` and use a standard colour in the Eclipse platform. That way, ColorThemese plugin will correctly update this one as well (of course, users can still change this to another color).

From the Eclipse extension point docs:

```
defaultsTo - the id of another color definition that is the default setting for the receiver. When there is no preference for this color the color registry will have the value of defaultsTo set for it in the registry. Only one of defaultsTo, value or colorFactory may be defined.
```

dragos:
Great, thanks! I think we could use some `defaultsTo` for some colors (maybe not all), in order to allow ColorTheme to pick the same background as for other editors or consoles.

dragos:
Otherwise LGTM

org.scala-ide.sdt.core/plugin.xml
sschaef: I didn't understand he difference between `defaultTo` and `value` when I read their description, now I don't understand what you mean by "colour in the Eclipse platform". I think the ColorThemes plugin just changes the values that belong to all existing preferences and therefore this plugin would also need a patch to support these values. Or do you mean that we should not use our own ids?

dragos: The difference is that `defaultsTo` falls back to a color ID, like the background color of the text editors. That means there's an additional level of indirection. I'm pretty sure ColorThemes doesn't change all colors, but just the ones that come from the platform and a few Java ones. We could patch ColorThemes, but if it works without, it's simpler.

sschaef:
I tried different things, but I'm unable to come up with a solution that uses the `defaultsTo` mechanism. The pref id of the editor background color should be `org.eclipse.ui.editors.backgroundColor` (or `org.eclipse.ui.internal.texteditor.ITextEditorThemeConstants.PREFERENCE_COLOR_BACKGROUND`) which is set to 0xFFFFFF by default. The problem is that when I change it in "General > Editors > Text Editors > Background color", it is persisted with a different id in a different preference store.

`org.eclipse.ui.internal.editors.text.TextEditorDefaultsPreferencePage` stores them in a `org.eclipse.ui.preferences.ScopedPreferenceStore` (whose scope id is `org.eclipse.ui.editors`) with the id `org.eclipse.ui.texteditor.AbstractTextEditor.PREFERENCE_COLOR_BACKGROUND`. If anyone knows how to configure the color preference to access a specific scope please speak up. I don't want to invest further time into this and merge it as is.

PR #925, opened 2015-04-14
[WIP] Check compiler plugins for version conflicts and cleanup

dragos:
PC options in case of incompatibilities.

Compiler plugins (like macro-paradise) are cross-compiled with
the full Scala version, and they are not binary compatible across
minor versions. We acknowledge that and check it properly in the
classpath validator.

Compiler plugins can’t be instantiated in the PC unless the
project Scala version (Installation) is exactly the same as
the platform (the one Eclipse is running). This commit fixes PC
crashes due to such incompatibilities.

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/core/internal/project/ClasspathManagement.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/core/internal/project/ClasspathManagement.scala
@@ -45,6 +45,7 @@ import scala.util.Success
 import org.scalaide.core.IScalaClasspath
 import org.scalaide.core.ScalaInstallationChange
 import org.eclipse.jface.preference.IPersistentPreferenceStore
+import scala.tools.nsc.Settings
 
 /** The Scala classpath broken down in the JDK, Scala library and user library.
  *
@@ -231,7 +232,7 @@ trait ClasspathManagement extends HasLogger { self: ScalaProject =>
       classpathCheckLock.synchronized {
         // mark as in progress
         classpathHasBeenChecked = false
-            checkClasspath()
+          checkClasspath()
         if (classpathValid) {
           // no point in resetting compilers on an invalid classpath,
           // it would not work anyway. But we need to reset them if the classpath
@@ -340,7 +341,7 @@ trait ClasspathManagement extends HasLogger { self: ScalaProject =>
     val errors =
       validateScalaLibrary(scalaLibraries, canFixInstallationFromScalaLib) ++
         (if (withVersionClasspathValidator) {
-          validateBinaryVersionsOnClasspath()
+          validateBinaryVersionsOnClasspath() ++ validateCompilerPlugins()
         } else {
           Seq()
         })
@@ -493,5 +494,42 @@ trait ClasspathManagement extends HasLogger { self: ScalaProject =>
     errors.toSeq
   }
 
+  /** Make sure compiler plugins are binary compatible with the selected Scala installation.
+   *
+   *  Unlike regular dependencies, compiler plugins must be cross-compiled with the exact compiler
+   *  version used for building.
+   */
+  private def validateCompilerPlugins(): Seq[(Int, String, String)] = if (!isUsingCompatibilityMode()) Seq() else {
+    val errors = mutable.ListBuffer[(Int, String, String)]()
+
+    def paths(p: String) =
+      p.split(File.pathSeparator).toSeq.filterNot(_.isEmpty).distinct
+
+    def scanDir(d: String) = {
+      val entries = Option(new File(d).listFiles).getOrElse(Array.empty[File])
+      for (f <- entries if f.isFile)
+        yield f.getAbsolutePath
+    }
+
+    val plugins = paths(storage.getString("Xpluginsdir")).flatMap(scanDir) ++ paths(storage.getString("Xplugin"))
+    val installation = effectiveScalaInstallation()
+    def error(v: ScalaVersion, path: String) =
+      (IMarker.SEVERITY_ERROR,
+       s"Compiler plugin ${new Path(path).lastSegment()} is cross-compiled with incompatible version for this project: ${v.unparse} vs ${installation.version.unparse}",
+       SdtConstants.ScalaVersionProblemMarkerId)
+
+    for (p <- plugins if p.nonEmpty) {
+      p match {
+        case VersionInFile(version) if (version != installation.version) =>
+          errors += error(version, p)
+        case _ =>
+          ()
+      }
+    }
+
+    errors.toSeq
+  }
+
   private class cpMarkerFactory(key:String) extends MarkerFactory(key)
+
 }
--- a/org.scala-ide.sdt.core/src/org/scalaide/core/internal/project/ScalaProject.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/core/internal/project/ScalaProject.scala
@@ -427,20 +427,27 @@ class ScalaProject private (val underlying: IProject) extends ClasspathManagemen
   /** Compiler settings that are honored by the presentation compiler. */
   private def isPCSetting(settings: Settings): Set[Settings#Setting] = {
     import settings.{ plugin => pluginSetting, _ }
-    Set(deprecation,
+
+    val compilerPluginSettings: Set[Settings#Setting] = Set(pluginOptions,
+      pluginSetting,
+      pluginsDir)
+
+    val generalSettings: Set[Settings#Setting] = Set(deprecation,
       unchecked,
-      pluginOptions,
       verbose,
       Xexperimental,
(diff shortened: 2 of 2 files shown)


org.scala-ide.sdt.core/src/org/scalaide/core/internal/project/ClasspathManagement.scala
```diff
@@ -231,7 +232,7 @@ trait ClasspathManagement extends HasLogger { self: ScalaProject =>
       classpathCheckLock.synchronized {
         // mark as in progress
         classpathHasBeenChecked = false
-            checkClasspath()
+          checkClasspath()
```
sschaef: indent is still wrong

org.scala-ide.sdt.core/src/org/scalaide/core/internal/project/ClasspathManagement.scala
```diff
@@ -493,5 +494,42 @@ trait ClasspathManagement extends HasLogger { self: ScalaProject =>
...
+  /** Make sure compiler plugins are binary compatible with the selected Scala installation.
+   *
+   *  Unlike regular dependencies, compiler plugins must be cross-compiled with the exact compiler
+   *  version used for building.
+   */
+  private def validateCompilerPlugins(): Seq[(Int, String, String)] = if (!isUsingCompatibilityMode()) Seq() else {
+    val errors = mutable.ListBuffer[(Int, String, String)]()
+
+    def paths(p: String) =
+      p.split(File.pathSeparator).toSeq.filterNot(_.isEmpty).distinct
```
dragos: Actually, as strange as this is, the separator is `,`, not path separator.

org.scala-ide.sdt.core/src/org/scalaide/core/internal/project/ClasspathManagement.scala
```diff
@@ -493,5 +494,42 @@ trait ClasspathManagement extends HasLogger { self: ScalaProject =>
...
+
+    def scanDir(d: String) = {
+      val entries = Option(new File(d).listFiles).getOrElse(Array.empty[File])
+      for (f <- entries if f.isFile)
+        yield f.getAbsolutePath
+    }
+
+    val plugins = paths(storage.getString("Xpluginsdir")).flatMap(scanDir) ++ paths(storage.getString("Xplugin"))
+    val installation = effectiveScalaInstallation()
```
sschaef: Why `effectiveScalaInstallation?` In the description you said that the only valid Scala version is the one of the IDE, so shouldn't we get this one instead?

org.scala-ide.sdt.core/src/org/scalaide/core/internal/project/ClasspathManagement.scala
```diff
@@ -493,5 +494,42 @@ trait ClasspathManagement extends HasLogger { self: ScalaProject =>
     errors.toSeq
   }
 
+  /** Make sure compiler plugins are binary compatible with the selected Scala installation.
+   *
+   *  Unlike regular dependencies, compiler plugins must be cross-compiled with the exact compiler
+   *  version used for building.
+   */
+  private def validateCompilerPlugins(): Seq[(Int, String, String)] = if (!isUsingCompatibilityMode()) Seq() else {
```
sschaef: I don't understand the compatibility mode or better what the difference between -Xsource and the IDE specific Scala installation is. Can users set both?

org.scala-ide.sdt.core/src/org/scalaide/core/internal/project/ClasspathManagement.scala
```diff
@@ -45,6 +45,7 @@ import scala.util.Success
 import org.scalaide.core.IScalaClasspath
 import org.scalaide.core.ScalaInstallationChange
 import org.eclipse.jface.preference.IPersistentPreferenceStore
+import scala.tools.nsc.Settings
```
Kwestor: I know it's not in scope of this PR, but those imports are terrible, could you please organize them?

org.scala-ide.sdt.core/src/org/scalaide/core/internal/project/ClasspathManagement.scala
Kwestor: Could you introduce some domain case class instead of this `Tuple3`?

org.scala-ide.sdt.core/src/org/scalaide/core/internal/project/ScalaProject.scala
```diff
@@ -427,20 +427,27 @@ class ScalaProject private (val underlying: IProject) extends ClasspathManagemen
...
       YpresentationDebug,
       YpresentationVerbose,
       YpresentationLog,
       YpresentationReplay,
       YpresentationDelay)
+
+    if (effectiveScalaInstallation().version == ScalaInstallation.platformInstallation.version)
+      generalSettings ++ compilerPluginSettings
+    else
+      generalSettings
```
sschaef: I don't understand why this solves the problem. I have a test project, which is configured to use 2.11.6 paradise plugin. But my IDE has version 2.11.7. With this change the PC errors are gone (semantic highlighting on the other side doesn't work anymore) but where does the 2.11.7 paradise plugin come from? `pluginsDir` is set to `-Xpluginsdir = misc/scala-devel/plugins`, where does this directory point to?

Should we always ignore user plugin preferences for the PC? If the wrong version of a compiler plugin is used and the PC is configured with this one, we will get error markers whenever the compiler plugin is used.

org.scala-ide.sdt.core/src/org/scalaide/core/internal/project/ClasspathManagement.scala
sschaef: fixed

org.scala-ide.sdt.core/src/org/scalaide/core/internal/project/ClasspathManagement.scala
dragos: This check is there for the builder compiler, not the presentation compiler. Compiler plugins would work fine when building if they really match the Scala compiler version used for building (say, `2.10.4`).

org.scala-ide.sdt.core/src/org/scalaide/core/internal/project/ClasspathManagement.scala
dragos: They are indeed linked. The `-Xsource` flag is telling the type-checker to emit the same errors as in that major revision of the compiler. It has no bearing on the build compiler. Setting a Scala installation from the 2.10 series will automatically add `-Xsource`.

Probably that test should not be there, in fact, since compiler plugins may be incompatible even when running with a 2.11 compiler.

dragos: Yeah, a case class would be nicer.

org.scala-ide.sdt.core/src/org/scalaide/core/internal/project/ScalaProject.scala
dragos: You're saying that the presentation compiler error markers are gone... so that's what should happen. In general, a compiler plugin that doesn't match the exact Scala compiler version can't be instantiated, so the presentation compiler will simply ignore it.

If the compiler plugin doesn't match the installation, the earlier check in `checkClasspath` will flag it as an error marker at *build time*. For building we can't skip compiler plugins, they are required for a correct output. So the user will have to set a correct Scala installation. This would fix the build. The presentation compiler will continue to skip the compiler plugin, since the PC is locked down on the platform installation.

The default path, `misc/scala-devel/plugins` is probably wrong, I don't know exactly where it comes from. Probably the Scala compiler itself.

org.scala-ide.sdt.core/src/org/scalaide/core/internal/project/ClasspathManagement.scala
sschaef: So, we can't run the PC in compatibility mode or with a specific Scala installation?

dragos: Compiler plugins can use internal methods, so they can't run inside the PC unless they match the exact version. The PC is always running the bundled Eclipse version (`platformInstallation`). `-Xsource` and compatibility mode are mostly about what the type-checker flags as errors: it will type-check "as-if" it was the 2.10 compiler, so any bugfixes in 2.11 that give different errors are behind the `-Xsource` flag.

However, the checks here are for for the build compiler. For building, we instantiate a `scala-compiler.jar` from the project Scala Installation, so if the compiler plugin matches that version, it should work.

Does this answer your question?

sschaef:
Superseded by #935

PR #927, opened 2015-04-17
Disable flaky tests that often fail in Scala PR validation

lrytz:
The tests in HotCodeReplaceTest and the test
PresentationCompilerActivityListenerTest#changingManyPreferencesAtOnce
fail often in Scala's PR validation.

Links to recent failures:

https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-ide/755/console
  - doNotDropLastFrame(org.scalaide.debug.internal.hcr.HotCodeReplaceTest)
  - hcrWithDisabledAutomaticDroppingFrames(org.scalaide.debug.internal.hcr.HotCodeReplaceTest)
  - successfulHcrWithSimpleMethod(org.scalaide.debug.internal.hcr.HotCodeReplaceTest)

https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-ide/753/console
  - hcrWithDisabledAutomaticDroppingFrames(org.scalaide.debug.internal.hcr.HotCodeReplaceTest)
  - successfulHcrWithSimpleMethod(org.scalaide.debug.internal.hcr.HotCodeReplaceTest)

https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-ide/752/console
  - prohibitedDroppingObsoleteFramesManuallyDoesNotAffectAutomaticDropping(org.scalaide.debug.internal.hcr.HotCodeReplaceTest)

https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-ide/751/console
  - successfulHcrWithMethodNotInStackTrace(org.scalaide.debug.internal.hcr.HotCodeReplaceTest)
  - successfulHcrWithSimpleMethod(org.scalaide.debug.internal.hcr.HotCodeReplaceTest)

https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-ide/750/console
  - changingManyPreferencesAtOnce(org.scalaide.core.compiler.PresentationCompilerActivityListenerTest)

https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-ide/749/console
  - prohibitedDroppingObsoleteFramesManuallyDoesNotAffectAutomaticDropping(org.scalaide.debug.internal.hcr.HotCodeReplaceTest)

https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-ide/748/console
  - changingManyPreferencesAtOnce(org.scalaide.core.compiler.PresentationCompilerActivityListenerTest)

https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-ide/746/console
  - prohibitedDroppingObsoleteFramesManuallyDoesNotAffectAutomaticDropping(org.scalaide.debug.internal.hcr.HotCodeReplaceTest)

https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-ide/737/console
  - changingManyPreferencesAtOnce(org.scalaide.core.compiler.PresentationCompilerActivityListenerTest)

https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-ide/736/console
  - changingManyPreferencesAtOnce(org.scalaide.core.compiler.PresentationCompilerActivityListenerTest)

### DIFF
--- a/org.scala-ide.sdt.core.tests/src/org/scalaide/core/compiler/PresentationCompilerActivityListenerTest.scala
+++ b/org.scala-ide.sdt.core.tests/src/org/scalaide/core/compiler/PresentationCompilerActivityListenerTest.scala
@@ -5,6 +5,7 @@ package org.scalaide.core.compiler
 
 import org.eclipse.jface.util.PropertyChangeEvent
 import org.junit.Test
+import org.junit.Ignore
 import org.junit.Assert._
 import org.scalaide.ui.internal.preferences.ResourcesPreferences
 import Thread.sleep
@@ -221,6 +222,7 @@ class PresentationCompilerActivityListenerTest {
     listener.stop()
   }
 
+  @Ignore("Flaky, often fails in Scala PR validation.")
   @Test
   def changingManyPreferencesAtOnce(): Unit = {
     val shutdownMock = new MockShutdownFun
--- a/org.scala-ide.sdt.debug.tests/src/org/scalaide/debug/internal/hcr/HotCodeReplaceTest.scala
+++ b/org.scala-ide.sdt.debug.tests/src/org/scalaide/debug/internal/hcr/HotCodeReplaceTest.scala
@@ -58,6 +58,7 @@ private object HotCodeReplaceTest {
  * Tests whether HCR works (classes are correctly replaced in VM and we get new values)
  * and whether associated settings are correctly applied.
  */
+@Ignore("Flaky, often fails in Scala PR validation.")
 class HotCodeReplaceTest
     extends TestProjectSetup("hot-code-replace", bundleName = "org.scala-ide.sdt.debug.tests")
     with ScalaDebugRunningTest {


lrytz:
review by @mpociecha @dragos

misto:
add to whitelist

mpociecha:
PresentationCompilerActivityListenerTest:
It was created quite long time ago and probably we never had problems with these tests. On the other hand i know that simple Thread.sleep is not ideal solution. Theoretically we can increase time limits, but In case of these tests it was important to have some actions done in time. It's a question whether increased timeout would guarantee that it will never fail in case of any slowdowns on Scala Jenkins.

@dragos WDYT?

HCRTest
It seems that also in this case time limits were not enough on Scala Jenkins - not all events were propagated or building the project and redefining classes didn't finish. I'll try to change it to use in tests some registered listener instead and don't rely on predefined timeouts so much.

lrytz:
Thanks for taking a look @mpociecha. We'd appreciate merging this PR first and fixing the tests later, so that our PR validation is unblocked.

mpociecha:
Yes, I know. Simply I can't merge it because of lack of permissions. I only can give you my LGTM.

Looking at Scala Jenkins, some new builds are green so I hope you're not completely blocked.

sschaef:
I'm not in favor of disabling the tests completely. We have a mechanism to retry them instead, we should first go with that solution, given that the tests fail only sometimes.

sschaef:
Superseded by #928

sschaef:
@lrytz All the mentioned tests are now marked as flaky. Please send me a notification once the tests fail nevertheless and I'll revert my commit and instead merge yours. We'll then have to look later what to do with these tests.

lrytz:
Thanks, @sschaef!

mpociecha:
@lrytz @sschaef Unfortunately it didn't help.

dragos:
There's another method in test utils, I think it's called `waitUntil`. You can find the way we use it in other places. I should have suggested that the first time I saw the `Thread.sleep`. It's faster because it doesn't wait for a fixed amount of time, but instead retries until either timeouts or the condition passes. You could use that instead, with a timeout of 10 seconds or more.

We should really check if there is no other way to handle this.

sschaef:
Which `Thread.sleep` do you mean?

mpociecha:
@dragos yes, I saw SDTTestUtils.waitUntil. I even use it in my current, modified version. Anyway just replacing Thread.sleep with it wouldn't be enough, since we have asynchronous processing and also events sent to actors without waiting for the response.

mpociecha:
In other words - we need to have something what we can easily check (whether it finished). Previously there were nothing like that. Therefore, I will wait for the notification:
- We register a subscriber like in the case of this one which shows dialogs about failures (this particular one is not registered during tests).
- There's added new HCRSucceeded type of the event.
- We send some messages to actors to refresh stack frames, breakpoints etc., and after that we notify subscribers (so this test one) - theoretically it's possible that in this short period we'll notify the subscriber before all events in actors will be processed. I don't think it makes sense to change ! to !?, and wait in HotCodeReplaceExecutor (so to block ScalaDebugTarget's companion actor) though.

mpociecha:
"It's faster because it doesn't wait for a fixed amount of time"
Yes, I know. On my computer the condition usually passes after the 2nd check (100 millis) - so we won't need to wait whole 500 millis like before.

dragos:
@mpociecha unfortunately the PR validation build is running on much more resource-constrained systems...

I still don't understand why you couldn't use `waitUntil` with the exact same condition that you use for the test to pass/succeed. Roughly, replace

```
currentFrameLocation mustEqual ClassMethodEnd
```

with

```
waitUntil(TIMEOUT) { currentFrameLocation == ClassMethodEnd }
currentFrameLocation mustEqual ClassMethodEnd // repeat the test, in case it timed-out, this will fail
```

Obviously, a better solution based on some event signalling is preferable, but this might be a simple temporary solution.

mpociecha:
@dragos 
"Obviously, a better solution based on some event signalling is preferable"
I fully agree - it's better to check something when we know it's proper moment to do that. It's introduced in https://github.com/scala-ide/scala-ide/pull/930. I hope it's enough for Scala's Jenkins even taking into account lack of events to check from BreakpointSupport (and that it doesn't matter how long everything lasts if it's < time limit).

Regarding your proposed solution:

1. Indeed, it would work in cases when we expect that after some time we automatically dropped frames - so there's new state which we can check. Then we also would have to check whether the test session is suspended.

2. When we changed some code and there's no stack frame containing the modified method, we have to:
 - drop frames manually
 - go to breakpoint once again
 - check that variables have new values

  We would have to repeat all these steps in each loop iteration - and I'm not sure whether it would be good idea.

3. In negative cases (we want to check that we didn't perform HCR so basically nothing changed) we don't have the value which we could observe. Therefore, there are still Thread.sleeps. I don't have better solution right now.
FYI, I tested them with changed settings to force failed tests. The current amount of millis is more than enough to fail not only on the computer which I usually use, but also on my other, 4-year-old laptop. In addition I opened few IDE's, ran few videos and some compilation in sbt to have ~100% CPU usage and ~0 free RAM => to test it in a bit more unusual conditions.

PR #929, opened 2015-04-17
Revert "Mark several tests as flaky"

sschaef:
@mpociecha @dragos This only disables the HCR tests, should this be increased to the PC tests mentioned in #927?


mpociecha:
It's hard to say whether marking `changingManyPreferencesAtOnce` as flaky solved the problem completely, since it was tested only during two builds. On the other hand I also don't like the idea of disabling tests completely.

@dragos WDYT?

sschaef:
@lrytz You can start another IDE validation build but I can't guarantee that it works. :(

lrytz:
OK, I started two more builds - thanks for the help!

PR #930, opened 2015-04-19
Smarter HCR-related tests

mpociecha:
So far there was Thread.sleep used with some hardcoded value to ensure
that all events were propagated correctly after the build. In general
it was:

a) buld project with changes
b) there are some post-build actions performed and events are sent in
a background so wait 500 millis

It was causing problems in case of Scala's Jenkins for which 500
millis was too low value. Also it's not the best idea to wait 500
millis (always) even if it's not needed.

Therefore, this commit provides a bit smarter solution. From now on
HCR-related publisher publishes the `HCRSucceeded` event, when HCR is
finished. In tests we take into account what happens under the hood
in aforementioned point b. It is:

1. perform HCR
2. send a message to the thread-related actor => some asynchronous
processing (at the end we fire an event)
3. send a message to the breakpoints-related actor => some
asynchronous processing
4. notify subscribers about finished HCR (asynchronously)

In tests we wait for the events from points 2 and 4 (with some quite
big time limit, what should be helpful in case of Jenkins), and after
that check whether current stack frame is correct, variables have
proper values and so on. Unfortunately there's no event etc. from
`BreakpointSupport` so we check nothing related to the third point.

From now on the test checking whether disabling HCR works correctly
checks only that `ScalaHotCodeReplaceManager` is not created for
ScalaDebugTarget. It's enough.

There's also added to `ScalaHotCodeReplaceManager` the detailed
description how HCR works in case of errors in Java or Scala code
(the differences are a result of differences in deltas which we get
in events).

### DIFF
--- a/org.scala-ide.sdt.core.tests/src/org/scalaide/core/testsetup/SDTTestUtils.scala
+++ b/org.scala-ide.sdt.core.tests/src/org/scalaide/core/testsetup/SDTTestUtils.scala
@@ -9,6 +9,7 @@ import java.io.ByteArrayInputStream
 import java.io.File
 import java.io.IOException
 import java.io.InputStream
+import java.util.concurrent.TimeoutException
 import org.eclipse.core.resources.IContainer
 import org.eclipse.core.resources.IFile
 import org.eclipse.core.resources.IFolder
@@ -254,13 +255,15 @@ object SDTTestUtils extends HasLogger {
   }
 
   /** Wait until `pred` is true, or timeout (in ms). */
-  def waitUntil(timeout: Int)(pred: => Boolean): Unit = {
+  def waitUntil(timeout: Int, withTimeoutException: Boolean = false)(pred: => Boolean): Unit = {
     val start = System.currentTimeMillis()
     var cond = pred
     while ((System.currentTimeMillis() < start + timeout) && !cond) {
       Thread.sleep(100)
       cond = pred
     }
+    if (!cond && withTimeoutException)
+      throw new TimeoutException(s"Predicate is not fulfiled after declared time limit ($timeout millis).")
   }
 
   /**
--- a/org.scala-ide.sdt.core.tests/src/org/scalaide/core/testsetup/TestProjectSetup.scala
+++ b/org.scala-ide.sdt.core.tests/src/org/scalaide/core/testsetup/TestProjectSetup.scala
@@ -1,6 +1,5 @@
 package org.scalaide.core.testsetup
 
-import scala.concurrent.duration._
 import scala.tools.nsc.interactive.Response
 
 import org.eclipse.core.resources.IFile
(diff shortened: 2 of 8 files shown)


mpociecha:
It would be nice if it would be possible to test it on Scala's Jenkins in non-intrusive way.

dragos:
This is a definite improvement. However, any `Thread.sleep` will eventually byte us. Couldn't we simply test some logic that determines whether HCR should be performed or not, and simply test that method? That would return a `Boolean`, and be synchronous. At least, that would be run right after a build (that we already have events for).

mpociecha:
First of all, I know it would be nice to have real unit tests - not only such complex integration ones. Just sometimes is hard to mock everything.

Regarding your remark:

1. Disabled HCR: `hcrManager` in `ScalaDebugTarget` should be None. Do you think it's enough to test this one condition? We could just:
 - ensure that the flag is set to false,
 - start the test session
 - check the value of `hcrManager` (None is expected)
 - and this even without making any changes in code, rebulding projects and so on.
2. Errors during compilation: It's not so simple. We need to add some line which doesn't compile to have error markers generated. Moreover we perform this check when traversing `IResourceDelta`, which we got from event, in `ChangedClassFilesVisitor`. Anyway some integration test is needed in this case, I think.

mpociecha:
@dragos 
Just in case: yes, I saw "At least, that would be run right after a build (that we already have events for).".

FYI, the second point is even something more than only a check whether HCR should be performed. When the source file related to the modified class contains an error, we just skip this particular class (similarly to the implementation for Java). HCR is still performed if there are any other remaining classes to replace.
I was thinking about this for a while and maybe it would be better to have these tests a bit more sophisticated => e.g. to have the second source file. Then we could:
 - change both files,
 - one of them contains an errors,
 - rebuild a project (HCR is performed so we can wait for events),
 - check that we dropped frames,
 - go to the breakpoint once again,
 - check that one value is changed and the second is not (due to the error).

Also I would get rid of the case for the method which is not in stack trace. It rather wouldn't add an additional information.

Together with the solution proposed for the disabled HCR it would allow us to remove naive Thread.sleep completely.

Let me know whether it sounds good to you.

dragos:
Sounds good!

dragos:
I only had a very quick look at the code, so if someone else could go through the code before merging that'd be great!

mpociecha:
Ok. I'll change it when I will have some spare time. And will ask Simon to take a look.

Kwestor:
I already had taken a look at it (internally, before PR was submitted), I don't know if it count ;)

mpociecha:
I pushed corrections.

The build succeeded but I don't understand one thing: I removed `@Ignore`, but it seems HCR-related tests were anyway skipped during the build (https://jenkins.scala-ide.org:8496/jenkins/job/ghprb-scala-ide-multi-conf/87/eclipse_platform=eclipse-luna/testReport/org.scalaide.debug.internal.hcr/).

dragos:
LGTM

mpociecha:
Thanks. I've just wanted to ask Simon to to take a look, but you were quicker. :) Now we'll see whether Scala's Jenkins will be happy about these changes.

sschaef:
Some of the tests now don't pass on my system (Java7+Java8).

mpociecha:
Can you send me details via mail? I tested it several times on two computers, and it seems to work on two different Jenkins instances.

mpociecha:
Scala's Jenkins once built it correctly but now it failed. Could we ignore this test class once again? Sorry for that. :/

sschaef:
See #934

PR #931, opened 2015-04-21
Fixes StackOverflowException in ScalaIndenter

sschaef:
The main problem is an exponential blowup of a regex. Another problem is
that the val `restOfTheLine` doesn't really contain the rest of the line,
but the start of the string literal until the offset, what in this case
is the end of the string literal. Another problem is that the regex is
too complicated.

The easiest way to fix the problem was to simplify the regex and to stop
the regex computation as soon as a newline character occurs, which
happens because `.` doesn't match newline characters.

No tests included because a test suite doesn't exist. The following
should be the behavior (^ is the cursor before/after `<enter>` is pressed):

    def f = ""^
    //==>
    def f = ""
    ^

and

    def f = ^""
    //==>
    def f =
      ^""

Fixes #1002433

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ScalaIndenter.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ScalaIndenter.scala
@@ -1517,7 +1517,7 @@ class ScalaIndenter(
    */
   private def isStringOrCharLiteralAssignment(referenceTokenPos: Int, offset: Int) = {
     val restOfTheLine = document.get(referenceTokenPos, Math.max(offset - referenceTokenPos, 0))
-    val charLit = "'.'"
+    val charLit = """'(\\?.|\\u[\da-fA-F]{4})'"""
     val stringLit = "\".*\""
     val regex = s"($charLit|$stringLit)".r
     regex.findFirstIn(restOfTheLine.trim).isDefined
--- a/org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ScalaIndenter.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ScalaIndenter.scala
@@ -534,7 +534,6 @@ class ScalaIndenter(
     val maxCopyLength = if (tabSize > 0) minLength - minLength % tabSize else minLength // maximum indent to copy
     stripExceedingChars(buffer, maxCopyLength)
 
-
     def safeDivision(num: Int, denom: Int): (Int,Int) = if (denom > 0) (num / denom, num % denom) else (0, num)
 
     // add additional indent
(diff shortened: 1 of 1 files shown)


org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ScalaIndenter.scala
```diff
@@ -1518,10 +1517,9 @@ class ScalaIndenter(
    */
   private def isStringOrCharLiteralAssignment(referenceTokenPos: Int, offset: Int) = {
     val restOfTheLine = document.get(referenceTokenPos, Math.max(offset - referenceTokenPos, 0))
-    val charLit = """'(.|\\\w|\\\\)'"""
-    val stringLit = """"(\\"|[^"])*")"""
-
-    val regex = s"(${charLit}|${stringLit}$$".r
+    val charLit = "'.'"
+    val stringLit = "\".*\""
```
dragos: I still think you need to accept escaped strings. This regex does not match `"quote inside \""`.

sschaef: It does. `.` eats everything.

dragos: I see, it relies on the longest match rule. So if there are two string literals it would match both of them (`"abc", "def"` matches the whole string). I'm not sure it's a problem. But I still don't understand how this code triggered a stack overflow, it doesn't seem to be recursive. What is the actual benefit of this change?

sschaef: The only important rule is that we do an indent when the caret is after the `=` but before the rhs. The content of the rhs shouldn't matter.

sschaef: About the regex:

```scala
// matches immediately
scala> "a"*51+"a" matches "(aa|aa)*"
res0: Boolean = true

// takes a long times to compute a result
scala> "a"*52+"a" matches "(aa|aa)*"
res1: Boolean = false

// Again, matches immediately
scala> "a"*52+"a" matches "(aa)*"
res2: Boolean = false
```
In this example "(aa|aa)*" is not optimized to "(aa)*" and because the Java Regex engine uses recursive backtracking which has [exponential running time](http://stackoverflow.com/questions/8887724/why-can-regular-expressions-have-an-exponential-running-time) and therefore do overflow at some point.

I think the previous regex overflowed, because the alternative of `\\"|[^"]` is recursively applied for every character (similar as in the above example). My new regex does not have `"\".*\""` doesn't have an alternative anymore and therefore should match sequentially.

dragos: Sounds good, ok!

dragos: Though I'm not sure this will work for `val x = '\n'`?

sschaef: Oh my, Unicode literals are also not working

sschaef:
I added a new commit, which fixes the escaped signs and uncode literals.

dragos:
LGTM

PR #934, opened 2015-04-24
Disable flaky tests that often fail in Scala PR validation

sschaef:
The tests were recently re-enabled, because they were improved. But they
are still not improved enough.

See:

https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-ide/821/console


mpociecha:
For future reference:
It seems that for some reason from time to time, randomly we get an event with deltas only for changed source files. Therefore, HCR isn't performed in such situations in tests and events are never fired. In a normal situation we have deltas both for source (we simply skip them) and class files.

I didn't manage to reproduce these problems on some computers (3 computers with Windows). On the other hand, they occur quite often on others (tested on Ubuntu - it would be rather strange, if OS would make any difference in this case though).

PR #933, opened 2015-04-24
Update tasks after build

Kwestor:
During each build tasks are updated for every new and/or updated file.

Fixes [#1002137](https://scala-ide-portfolio.assembla.com/spaces/scala-ide/tickets/1002137#/activity/ticket:)

Lastly, the below disclaimer is required by the lawyers:

THE FOLLOWING DISCLAIMER APPLIES TO ALL SOFTWARE CODE AND OTHER MATERIALS CONTRIBUTED IN CONNECTION WITH THIS SOFTWARE:
THIS SOFTWARE IS LICENSED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS “AS IS” AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE AND ANY WARRANTY OF NON-INFRINGEMENT, ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE. THIS SOFTWARE MAY BE REDISTRIBUTED TO OTHERS ONLY BY EFFECTIVELY USING THIS OR ANOTHER EQUIVALENT DISCLAIMER IN ADDITION TO ANY OTHER REQUIRED LICENSE TERMS.
[...]

PR #936, opened 2015-04-29
Disable flaky test in ProjectDependenciesTest

sschaef:
This test failed a lot of times lately, see:

https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-ide/837/console

https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-ide/835/console

https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-ide/829/console

https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-ide/827/console

https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-ide/823/console

https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-ide/816/console

----

It is weird that it failed only in the last days, did anything change that made this test flaky?

### DIFF
--- a/org.scala-ide.sdt.core.tests/src/org/scalaide/core/sbtbuilder/ProjectDependenciesTest.scala
+++ b/org.scala-ide.sdt.core.tests/src/org/scalaide/core/sbtbuilder/ProjectDependenciesTest.scala
@@ -1,6 +1,7 @@
 package org.scalaide.core
 package sbtbuilder
 
+import org.junit.Ignore
 import org.junit.Test
 import org.eclipse.jdt.core.IJavaProject
 import org.eclipse.jdt.core.IClasspathEntry
@@ -98,6 +99,7 @@ class ProjectDependenciesTest {
     }
   }
 
+  @Ignore
   @Test def transitive_dep_indirect() {
     val Seq(prjA, prjB, prjC) = createProjects("A", "B", "C")
 


mpociecha:
Yes, I've also seen these failing builds. LGTM

mpociecha:
Looking only at dates, it started failing after merging https://github.com/scala-ide/scala-ide/pull/933, but maybe is it caused just by some changes in the CI configuration, if there were any? (I mean some new, specific state on Jenkins)?

mpociecha:
By the way, maybe would it be possible to have such tests marked in that way that they would be ignored/skipped (I assume that temporarily) only at Scala's Jenkins (where they are causing problems), but to still run them on dev machines and on Scala IDE's Jenkins?

Some env variable or something like that.

mpociecha:
Obviously it's a question whether Scala's Jenkins simply discovers some corner cases which should be fixed at some point.

Kwestor:
That commit introduced some delays into build process (looking into all comments for TODOs). How can I check what exactly failed (which assertion) on Scala Jenkins?

Btw - after my changes tests from `ProjectDependenciesTest` do not run at all on Windows, as deleting projects after test fails with "could not delete file". I have no idea how changing build times prevents one from deleting files.

org.scala-ide.sdt.core.tests/src/org/scalaide/core/sbtbuilder/ProjectDependenciesTest.scala
```diff
@@ -98,6 +99,7 @@ class ProjectDependenciesTest {
     }
   }
 
+  @Ignore
```
Kwestor: Could we mark it as flaky instead?

mpociecha: Lately they always fail (there was no build when it didn't fail, I think). I understood that it's the reason.

dragos: So, if it fails everytime, how do we know it's not a real regression?

mpociecha: I meant "on Scala's Jekins".

It simply passes on Scala IDE's Jenkins and when someone runs tests on his/her computer.

Except that I don't know. // I'm just reading these comments since I'm getting notifications (not that I was investigating this issue or something like that).

dragos:
The interesting thing is that the test does not fail, it errors (meaning an exception was thrown). I'm trying to look through the swaths of logs for a hint on what went wrong

lrytz:
Some details how it's related to #933: https://github.com/scala-ide/scala-ide/pull/933#issuecomment-98633087

dragos:
@lrytz thanks for digging out the stack trace, I don't know how I missed it. I guess we can close this PR now?

lrytz:
It looks like it should be good – I can't say right now because our jenkins slaves won't start :) It's a small fix (upgrade the default java due to https://jenkins-ci.org/content/good-bye-java6), but I don't all the tools to get it into our infra.

mpociecha:
Let's wait for another few builds. Anyway for sure it's better - https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-ide/871/

lrytz:
I'm fine closing this, since the build passed. The test was not flaky.

PR #937, opened 2015-04-30
Handle `DocDef` trees in the JUnit test finder.

dragos:
I added a few more lines of logging, in case this is not the only regression.

Fix #1002442

It'd be great to have this in 4.1 as well.


Kwestor:
LGTM

Btw - packa in Polish: https://www.google.pl/search?q=packa+na+muchy&tbm=isch ;)

dragos:
:laughing:

org.scala-ide.sdt.core/src/org/scalaide/core/internal/launching/JUnit4TestClassesCollector.scala
```diff
@@ -35,6 +35,7 @@ private[launching] abstract class JUnit4TestClassesCollector extends HasLogger {
 
     override def traverse(tree: Tree): Unit = tree match {
       case _: PackageDef => super.traverse(tree)
+      case _: DocDef => super.traverse(tree)
```
sschaef: Oh my, the Scala trees are always good for a surprise.

PR #938, opened 2015-04-30
Workaround for problems in conditional breakpoints

Kwestor:
`SPCIntegration` uses global debug state (from `ScalaDebugger`) which is
unavailable during conditional brakpoint check - condition is evaluated
**before** breakpoint is hit.

This commit wraps login in `try/catch` and ignores those errors. This
unfortunately means imports from debugged file will not be added before
expression and limited generics support.

Next week we will work on more permanent solution.


dragos:
LGTM

PR #939, opened 2015-04-30
Update to the new repository for SBT builds.

dragos:
The Typesafe artifactory repository is no longer available. We moved
existing artifacts to another repository. This should fix the build for everyone

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -91,7 +91,7 @@
     <!-- the repos containing the Scala dependencies -->
     <repo.scala-refactoring>${repo.scala-ide.root}/scala-refactoring-${scala.short.version}x</repo.scala-refactoring>
     <repo.scalariform>${repo.scala-ide.root}/scalariform-${scala.short.version}x</repo.scalariform>
-    <repo.typesafe>http://private-repo.typesafe.com/typesafe/ide-${scala.minor.version}</repo.typesafe>
+    <repo.typesafe>https://proxy-ch.typesafe.com:8082/artifactory/ide-${scala.minor.version}</repo.typesafe>
 
   </properties>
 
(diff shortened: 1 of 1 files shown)


sschaef:
It does still not work on a fresh maven repo.

dragos:
It probably does. You just need to wipe it, or add `-U` if you already tried it today. Maven caches the information that a snapshot *wasn't* available today, and it won't check again unless you force it with `-U`.

sschaef:
I tried on a complete fresh repo (with the `-Dmaven.repo.local=/tmp/repo` flag) where nothing was cached.

sschaef:
Would be nice if someone else could try that, not that in the end my maven is broken - wouldn't be the first time.

dragos:
I just tried it, it worked for me. Are you sure you didn't use the same temporary repo before?

mpociecha:
@sschaef  Okay, I'll try it in few minutes.

mpociecha:
For me Iulian's branch in connection with a clean repo builds without problem.

dragos:
Ok, I'll proceed with a merge. :)

sschaef:
Ok, I finally also succeeded in building it successfully. The problem was that I ran `mvn clean verify`, which did not work. `mvn clean install` in contrast led to success. I always thought `install` is just the phase after `verify` and therefore is doing additional stuff. How wrong I was.

dragos:
You shouldn't run `mvn` directly. Why not go through `build-all.sh`?

There's more than one project that gets built in `build-all.sh`, and the main build step needs to find those projects in your local maven repository. If you don't `install` them, there's no way to find them (unless you have an older version lying around)

sschaef:
Yeah, after thinking about it again yesterday evening, it occurred to me that the error message I saw is completely logical. So, `verify` is indeed doing what it should do. I have my own build scripts to make my live easier (building multiple repos at once etc.), therefore I do some things a little bit differently. Normally that works fine, with exception of the rare cases where it turns out that you fully need to understand how the build works.

PR #940, opened 2015-04-30
Typo and more accurate log message.

dragos:
..I can't think of a better description than this. :)

### DIFF
--- a/org.scala-ide.sdt.core.tests/test-workspace/completion/src/ticket_1000412/test/TestJavaExtends.java
+++ b/org.scala-ide.sdt.core.tests/test-workspace/completion/src/ticket_1000412/test/TestJavaExtends.java
@@ -28,7 +28,7 @@ public class TestJavaExtends extends ClassA {
         x=ge/*!*/
     }
 
-    privite void bar5() {
+    private void bar5() {
         ge/*!*/;
     }
 
--- a/org.scala-ide.sdt.core/src/org/scalaide/core/internal/builder/zinc/SbtBuildReporter.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/core/internal/builder/zinc/SbtBuildReporter.scala
@@ -87,7 +87,7 @@ private[zinc] class SbtBuildReporter(project: IScalaProject) extends xsbti.Repor
       val markerPos = MarkerFactory.RegionPosition(offset, identifierLength(pos.lineContent, pos.pointer), line)
       BuildProblemMarker.create(resource, severity, msg, markerPos)
     } else
-      logger.info("suppressed error in Java file: %s".format(msg))
+      logger.error(s"suppressed error in Java file ${resource.getFullPath}:$line: $msg")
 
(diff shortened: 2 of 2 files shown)


org.scala-ide.sdt.core.tests/test-workspace/completion/src/ticket_1000412/test/TestJavaExtends.java
```diff
@@ -28,7 +28,7 @@ private void bar4() {
         x=ge/*!*/
     }
 
-    privite void bar5() {
+    private void bar5() {
```
mpociecha: Hm, how were related tests working?

mpociecha:
Do you think that `severity` can be helpful? You could log also it.

sschaef:
May this be the reason for #936?

org.scala-ide.sdt.core/src/org/scalaide/core/internal/builder/zinc/SbtBuildReporter.scala
```diff
@@ -87,7 +87,7 @@ private[zinc] class SbtBuildReporter(project: IScalaProject) extends xsbti.Repor
       val markerPos = MarkerFactory.RegionPosition(offset, identifierLength(pos.lineContent, pos.pointer), line)
       BuildProblemMarker.create(resource, severity, msg, markerPos)
     } else
-      logger.info("suppressed error in Java file: %s".format(msg))
+      logger.info(s"suppressed error in Java file ${resource.getFullPath}:$line: $msg")
```
sschaef: It is an error, why not log it as such? Do we have another logger for Java errors?

dragos:
I don't think it's the cause of #936 :(

org.scala-ide.sdt.core/src/org/scalaide/core/internal/builder/zinc/SbtBuildReporter.scala
dragos: I agree we need to do something better here, but I don't have a good solution right now. Here's why we don't report it: Java sources are compiled by the Eclipse Java compiler. It will report its own error markers, later -- we'd be duplicating those.

Here's why we should report it: the Scala compiler reads those Java files because it needs to compile Scala sources. If there's an error in a Java file it *won't produce bytecode*. That's not a problem, as usually the Eclipse Java compiler already reported those errors. Now the beauty of this is that the Scala parser for Java files is a bit different from the Java one. You might have a correct Java file that leads to no errors (Eclipse Java thinks it's ok, the Scala compiler does not -- and won't compile Scala sources either) *and no bytecode at all*.

It's a catch-22.

dragos: Oh, I misread your comment. I thought you asked why we don't create a marker. Yes, it makes sense. I'll change the error level.

sschaef:
Ok, LGTM then.

dragos:
retest this please

PR #941, opened 2015-05-03
Fix compiler warning

mlangc:
Fixes #1002444

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/refactoring/internal/move/MoveClass.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/refactoring/internal/move/MoveClass.scala
@@ -131,6 +131,8 @@ class MoveClass extends RefactoringExecutorWithWizard {
             (change :: textChanges, newFiles)
           case (change: NewFileChange, (textChanges, newFilesChanges)) =>
             (textChanges, change :: newFilesChanges)
+          case (unexpected, _) =>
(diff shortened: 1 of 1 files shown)


org.scala-ide.sdt.core/src/org/scalaide/refactoring/internal/move/MoveClass.scala
```diff
@@ -131,6 +131,7 @@ class MoveClass extends RefactoringExecutorWithWizard {
             (change :: textChanges, newFiles)
           case (change: NewFileChange, (textChanges, newFilesChanges)) =>
             (textChanges, change :: newFilesChanges)
+          case (other, _) => throw new AssertionError(s"Unexpected change $other")
```
sschaef: Please improve the message in a way that it explains that only `TextChange` and `NewFileChange` make sense here.

dragos: I don't see any difference between throwing `MatchError` and `AssertionError`. Actually, as @sschaef noticed, `MatchError` gives you more information. I understand the warning is gone, but then let's report the error in a better way. (or use `@unchecked` to silence the warning).

mlangc:
Hmm, please tell me if you like the new message better; if not I'd go for the `@unchecked` option.

sschaef:
LGTM

PR #944, opened 2015-05-07
Refactor/shorter logging in ee

Kwestor:
First commit adds `TRACE` level to logger, second changes long log messages in expression evaluator to use it.

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/core/internal/logging/EclipseLogger.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/core/internal/logging/EclipseLogger.scala
@@ -33,47 +33,55 @@ object EclipseLogger extends Logger {
 
   private val lastCrash: AtomicReference[Throwable] = new AtomicReference
 
-  def debug(message: => Any) {
+  def trace(message: => Any): Unit = {
     info(message)
   }
 
-  def debug(message: => Any, t: Throwable) {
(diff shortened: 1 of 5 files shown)


org.scala-ide.sdt.core/src/org/scalaide/core/internal/logging/EclipseLogger.scala
```diff
@@ -33,47 +33,55 @@ object EclipseLogger extends Logger {
 
   private val lastCrash: AtomicReference[Throwable] = new AtomicReference
 
-  def debug(message: => Any) {
+  def trace(message: => Any): Unit = {
```
mpociecha: I would make them one-liners. ;)

Kwestor: My too, but I just copy-pasted code below.

dragos:
in the end it doesn't seem to make much of a difference... I guess the bulk comes from somewhere else

Kwestor:
Maybe from project compilation? We have a lot of test projects there and they are compiled before each test class. We have a change that integrates it into single compilation before **all** tests and then clean after, but it needs to be rebased and polished a bit. @romanowski will take care of in some free time.

Kwestor:
@dragos should we merge it anyway?

Kwestor:
Merging, as it does not hurt us to have some logs on `TRACE`.

PR #946, opened 2015-05-13
Fix migration preference initializer

Kwestor:
Runs initialization code only if workspace is present.

Fix #1002441


sschaef:
LGTM

PR #950, opened 2015-05-15
Move key binding for expression evaluator to different key

sschaef:
CTRL-SHIFT-Z invokes already the redo functionality.

Other key bindings that are mentioned in the ticket are not changed
because they don't seem to override anything.

Fixes #1002459

### DIFF
--- a/org.scala-ide.sdt.debug.expression/plugin.xml
+++ b/org.scala-ide.sdt.debug.expression/plugin.xml
@@ -35,7 +35,7 @@
             commandId="org.scalaide.debug.handler.RunSelectionInExpressionEvaluator"
             schemeId="org.eclipse.ui.defaultAcceleratorConfiguration"
             contextId="scala.tools.eclipse.scalaEditorScope"
-            sequence="M1+M2+Z"/>
+            sequence="M1+M2+K"/>
     </extension>
 	
     <extension


Kwestor:
On windows "redo" is on `CTRL+Y`, I'm curious which other shortcuts depends on platform.

pkukielka:
Would it be possible to use `CTRL + ALT + E`?
`E` is easier to associate with Expression Evaluator and it's possible to hit with one hand.
I didn't noticed anything under this binding (but correct me if I'm wrong).

sschaef:
I thought `CTRL+ALT` is difficult to type with two hands.

mpociecha:
Maybe. But it's easy with one hand. :)

mpociecha:
By the way, as Jerzy said, on Windows shortcuts are slightly different. Ctrl + Shift + Z was unassigned and was similar to Ctrl + Shift + X which opens Scala interpreter. That's why I chose this one.

sschaef:
@wpopielarski You are developing on Windows, could you please have a look if both `CTRL + ALT + E` and `CTRL+SHIFT+K` are free?

Kwestor:
I'm on Windows, both are free.

sschaef:
It doesn't really matter which key binding is used as long as it is free. The important thing is that users don't get interrupted through new features they may never need. It someone doesn't enjoy the default key binding, they can change it.

PR #953, opened 2015-05-18
Force quality code

sschaef:
This PR fixes all warnings in our codebase and enables a lot of scalac flags, including `-Xfatal-warning`, in our build in order to guarantee that in future no new warnings are introduced into the codebase. Beside from the build, the compiler options are also enabled in the project preferences and therefore directly lead to compilation errors in Eclipse.

The following warnings are enabled:
```
-feature
-unchecked
-deprecation
-language:_
-Xlint
-Xfatal-warnings
-Yno-adapted-args
-Ywarn-dead-code
-Ywarn-unused-import
-Ywarn-unused
```

I don't know what is the best way to review this PR - probably you need to review every commit separately. Otherwise the changes are too huge. Most changes are cosmetic but I also fixed a lot of deprecation warnings which are more unsafe.

I also introduced the type `org.scalaide.util.internal.Suppress` which allows us to suppress deprecation warnings (what a bad hack, it is a surprise that it is even possible). This is not a very nice solution but some deprecations either couldn't be fixed (Scala Actors) or I had no idea how to do it.

There is only one small downside that comes from enabling `-deprecation`: Whenever the API of scalac changes (or any other API we compile nightly) and a deprecation warning is introduced to a definition we use, our build breaks (and therefore also the `scala/scala` build). But this is only a small downside in my opinion, given that it is easy to alias a definition in the `Suppress` type and then rebuild `scala/scala`.


Kwestor:
Changes looks OK, I would like to look at them in the IDE, unfortunately without `Unit` commit nothing compiles. Maybe I will check original branch in IDE.

sschaef:
Oh dammit, I forgot to change the unit thing in the Eclipse preferences. Wait a second.

Kwestor:
No problem, I looked at previous branch.

There is one lat warning in our codebase: `Avoid block imports` from Scalastyle - do we intend to change it to `error`?

sschaef:
We didn't mark it as error because there were false positives in the Scalastyle check. We could rethink that though, I can't remember which cases were false positives.

sschaef:
There are merge errors now, but the code should compile in Eclipse and maven.

Kwestor:
I looked at cases reported in our codebase and it seems all of them are fixable right now.

Kwestor:
LGTM

sschaef:
@dragos Do you want to have a closer look to this PR?

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ScalaSourceFileEditor.scala
```diff
@@ -291,10 +290,12 @@ object ScalaSourceFileEditor {
...
   /** A thread-safe object for keeping track of Java reconciling listeners.*/
   private class ReconcilingListeners extends IJavaReconcilingListener {
-    private val reconcilingListeners = new ArrayBuffer[IJavaReconcilingListener] with SynchronizedBuffer[IJavaReconcilingListener]
+    import java.util.concurrent.ConcurrentLinkedQueue
+
+    private val reconcilingListeners = new ConcurrentLinkedQueue[IJavaReconcilingListener]()
 
     /** Return a snapshot of the currently registered `reconcilingListeners`. This is useful to avoid concurrency hazards when iterating on the `reconcilingListeners`. */
-    private def currentReconcilingListeners: List[IJavaReconcilingListener] = reconcilingListeners.toList
+    private def currentReconcilingListeners = reconcilingListeners.toArray.asInstanceOf[Array[IJavaReconcilingListener]]
```
sschaef: This produces a ClassCastException. Oh dear, how much I hate arrays.

Kwestor: btw - what was a warning there?

sschaef: `SynchronizedBuffer` is deprecated and I replaced it with `ConcurrentLinkedQueue`

Kwestor: Maybe use `scala.collection.JavaConverters` and `asScala`? It returns `Iterable[A]` in my micro-test in console.

dragos:
I'll have a quick look after today's meeting.

Kwestor:
I removed multi-imports in #959.

PR #954, opened 2015-05-19
Add Unit return type to procedures

sschaef:
Rebased on #953 - only the last commits are new.

It adds Unit return types to all procedures and enables them during the build by enabling the `-Xfuture` flag.


Kwestor:
LGTM

PR #957, opened 2015-05-26
Make codebase compatible with Java6

sschaef:
The constructor of AssertionError that takes two parameters was
introduced in Java7. The constructor of IllegalStateException however is
available in Java6 and should fit all our needs.

------

This PR is needed, since uber-build runs only with Java6 and after the introduction of `-Xfatal-warnings`, the code did no longer compile. The only reason why the code compiled at all, was because the compiler implicitly converted the two parameters to a `Tuple2` and passed it to a constructor that expected a single parameter of type `Any`: *facepalm*


adriaanm:
Autotupling strikes again! Good catch!

PR #958, opened 2015-05-27
Add description to "Create Scala Project" wizard

lancewf:
Fix #1002430


sschaef:
LGTM

sschaef:
Thanks for this patch! Hopefully we see further patches from your side...

PR #959, opened 2015-05-27
Last warnings removed from Scala files

Kwestor:
Using multi-imports is now an error (like all other warnings).

All cases of multi-imports organized.


sschaef:
Great! LGTM

PR #961, opened 2015-05-27
Set UTF-8 as default charset in test workspaces

Kwestor:
This is required as otherwise on Windows default `CP-1250` is used and some
tests (CompletionTests in particular) fails.

I tested it on my machine and all tests pass both in maven and in eclipse.


sschaef:
LGTM

PR #962, opened 2015-05-28
Refactor/remove todos

Kwestor:
As task list works now I took a look at some of those `TODO`s and `FIXME`s and fixed/removed them.

There are still 130+ there to look at, but most of them are really cryptic or they require some in-depth knowledge or a lot of work.


sschaef:
LGTM

I was not aware that we have that many task tags. Anyway it can only get better.

sschaef:
If we would have a source code quality score, it would have increased a little bit - once again.

PR #965, opened 2015-05-29
Commented code removed

Kwestor:
As commented code is essentially dead it is removed now unless there is a
comment explaining why it is commented out.

`TestScalaIndenter` uncommented, ignored with message.

If you know any reason to keep some of those deleted snippets, I'll update code with proper description.


Kwestor:
Oh, those were tests that only work from eclipse, not from maven.

sschaef:
`TestScalaIndenter` triggers an eclipse bundle that requires GUI support, which is not available on the server. I once worked on a migration of those, but the result wasn't very nice at all (see https://github.com/sschaef/scala-ide/tree/t-refactor-indent-test-suite5). For now, just ignore the whole test class.

sschaef:
LGTM

PR #967, opened 2015-06-03
Synchronize all calls into the JDI

wpopielarski:
JDI calls seem to be thread unsafe. To fix it all potential invocations
with side effects to JDI must be found and synchronized.

There are four things done:
 - JdiEntryGuardianAspect is a tool which scans all usages of
   com.sun.jdi.* invocation. This aspect needs to be deactivated
   in post review rework
 - JdiInvocationSynchronizer just sync the calls to JDI
 - jdiSynchronized is scala object which wraps JdiInvocationSynchronizer

In scala-ide.log can be found potential places requiring of the sync
discoved by Guardian aspect.

Fix #1002160


org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/ScalaSourceLookupParticipant.scala
```diff
@@ -12,10 +13,10 @@ object ScalaSourceLookupParticipant extends AbstractSourceLookupParticipant {
   def getSourceName(obj: AnyRef): String = {
     obj match {
       case stackFrame: ScalaStackFrame =>
-        stackFrame.getSourcePath
+         jdiSynchronized { stackFrame.getSourcePath }
```
wpopielarski: not sure if we really need a sync here

dragos:
Hm, so the aspect is only a debugging tool? Cool!

wpopielarski:
Yes, aspect is just a debugging tool to trace potential sync points.
Continuing, I'm close to rethink locking and do not create a separate
dedicated actor for it (explanation why is in scala doc). Instead I'm
testing an idea of Futures in actor code to move the lock out an actor
thread. More results are expected tomorrow. If it go off well I start the
conversion to akka actors (found paper how to do it painlessly or the least
pain :) ).

2015-06-14 12:09 GMT+02:00 Iulian Dragos <notifications@github.com>:

wpopielarski:
Guy,
ready for inspection. Please verify if there is still any place which requires synch on JDI call. In aspects aop.xml is a tool which I hope is helpful to nail potentials.

org.scala-ide.sdt.aspects/META-INF/aop.xml
```diff
@@ -31,5 +31,6 @@
 <aspect name="scala.tools.eclipse.contribution.weaving.jdt.debug.JDIModelPresentationAspect"/>
 <aspect name="scala.tools.eclipse.contribution.weaving.pde.ui.PluginImportAspect"/>
 <aspect name="scala.tools.eclipse.contribution.weaving.jdt.launching.JUnitLaunchConfigurationTabAspect"/>
+<!-- aspect name="scala.tools.eclipse.contribution.weaving.jdt.jdi.JdiEntryGuardianAspect"/ -->
```
dragos: I'm surprised this one isn't needed. Does it really apply the aspect when this line is commented? If so, just remove the line, no need to have stray comments here.

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/model/ScalaDebugCache.scala
```diff
@@ -275,16 +318,7 @@ protected[debug] class ScalaDebugCacheActor(debugCache: ScalaDebugCache, debugTa
...
-    nestedTypesCache.get(outerTypeName) match {
-      case Some(cache) =>
-        cache.types
-      case None =>
-        initializedRequestsAndCache(outerTypeName).types
-    }
-  }
-
-  private def initializedRequestsAndCache(outerTypeName: String): NestedTypesCache = {
+  private def findNestedTypesWithJdi(outerTypeName: String) = {
```
dragos: I'd rather put `jdiSynchronized` inside this method, rather than force all callers to call `jdiSynchronized` before calling this method.

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/model/ScalaDebugCache.scala
```diff
@@ -244,15 +265,37 @@ protected[debug] class ScalaDebugCacheActor(debugCache: ScalaDebugCache, debugTa
...
+      }
+      findCache(outerTypeName).onSuccess {
+        case (typeName, cache) =>
+          ScalaDebugCacheActor.this ! LoadedNestedTypesUpdate(typeName, cache)
+      }
     case AddClassPrepareEventListener(actor, outerTypeName) =>
-      addClassPreparedEventListener(actor, outerTypeName)
-      reply(true)
+      val replyTo = sender
+      become {
```
dragos: I don't fully understand the code changes here. Why did you need to create a new message and send it back to self in this way?

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/model/ScalaJdiEventDispatcher.scala
```diff
@@ -156,8 +166,15 @@ private class ScalaJdiEventDispatcherActor private (scalaDebugTargetActor: Suppr
         case result: Boolean => staySuspended |= result
       }
     }.andThen {
-      try if (!staySuspended) eventSet.resume()
-      finally ScalaJdiEventDispatcherActor.this ! FutureComputed
+      val resume = !staySuspended
+      val replyTo = this
+      Future {
+        jdiSynchronized {
```
dragos: I'm not sure this is actually necessary, but I guess the aspect is very pick and expects every call in `com.sun.jdi` to be synched...

org.scala-ide.sdt.aspects/META-INF/aop.xml
wpopielarski: uncomment it if you need to see all places where JDI is called. They are visible in scala-ide.log file

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/model/ScalaJdiEventDispatcher.scala
wpopielarski: ok

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/model/ScalaDebugCache.scala
wpopielarski: there is a problem that unfortunately actor's thread can be killed in any moment when info about types is retrieved and cached and then the sync lock is not released (I didn't find the way how to release lock when scala actor dies). So I had to push out this JDI operation to Future (outside actor's thread). This is connected also to your second question about 'findNestedTypeInJdi'. This method because of reason about needs to be called in Future thread. I hope I'm clear.

org.scala-ide.sdt.core/src/org/scalaide/util/Utils.scala
```diff
@@ -46,7 +50,37 @@ object Utils extends HasLogger {
...
+  /**
+   * Scala wrapper on [[JdiInvocationSynchronizer]]. Remark: the actor thread can be killed in any moment
+   * by [[ActorSystem]] so the lock is not released then. To avoid it call [[jdiSynchronized.apply]] wrapped
+   * in [[Future]] like in example:
+   * {{{
+   *   def receive = {
+   *     case msg =>
+   *       val replyTo = sender
+   *       Future {
+   *         jdiSynchronized {
```
pkukielka: Couldn't you move Future inside jdiSynchronized? This way user won't be able to make mistake and use it without Future:

```
object jdiSynchronized {
    def apply[T, R](code: => T)(fin: => R): Unit = {
        Future { JdiInvocationSynchronizer.instance.runSynchronized(callable(() => code)) } onComplete { fin }
    }
}
```

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/model/JdiRequestFactory.scala
```diff
@@ -54,23 +55,24 @@ object JdiRequestFactory {
...
+          breakpointRequest
+      }
     }
-  }
 
   /**
    * Create thread start request
    */
-  def createThreadStartRequest(virtualMachine: VirtualMachine): ThreadStartRequest = {
+  def createThreadStartRequest(virtualMachine: VirtualMachine): ThreadStartRequest = jdiSynchronized {
```
pkukielka: Are you sure none of that methods is called inside actor?
Because in that case you would need to pack it into Future.
Just asking :)

org.scala-ide.sdt.aspects/src/scala/tools/eclipse/contribution/weaving/jdt/jdi/JdiInvocationSynchronizer.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+public final class JdiInvocationSynchronizer {
+  private final static JdiInvocationSynchronizer INSTANCE = new JdiInvocationSynchronizer();
+  public final static long DEFAULT_TIMEOUT_IN_SEC = 10;
+
+  private JdiInvocationSynchronizer() {
+    // no need to instantiate outside
+  }
+
+  private final ReentrantLock jdiLock = new ReentrantLock();
```
dragos: On second thought: why not use `synchronized`? Synchronized blocks are guaranteed to be released, even if exceptions are thrown, so maybe that would simplify (also thinking about the actor code, and Futures that were needed for releasing the lock)

wpopielarski: A little bit defensive programming. A lock is tried to be retrieved, <code>synchronized</code> is not so wise I guess.

dragos: I still don't understand. `synchronized` makes sure that locks are released on all code paths. A manual lock requires manual intervention to ensure it's released. It's more versatile (if the locking path is not easily enclosed in a block, or if you want to pass the lock around to other threads), but I am not sure this is needed here.

wpopielarski: ok, I try if `synchronized` is good enough, just a KISS :)

sschaef:
LGTM. I used it a little bit in the debugger and didn't see a single `Couldn't retrieve variable's value` message, but I have no idea how to make it appear anyway. Let's see if more testing in the next days reveals more information

sschaef:
For now I think it is better to keep `JdiEntryGuardianAspect`, we can remove it in future if it turns out that the bug is really fixed.

wpopielarski:
thanks S

2015-08-07 0:10 GMT+02:00 Simon Schäfer <notifications@github.com>:

PR #970, opened 2015-06-18
Use OverlayPreferenceStore to handle correct UI semantics

sschaef:
In the preference page, there are several buttons that configure how any
chances should be propagated to the underlying preference store. When
Cancel is clicked, no changes should be propagated. When Apply and Save
are both clicked, the changes should only be propagated once. We use the
predefined class `OverlayPreferenceStore` to handle these semantics. It
only writes back its key-value pairs when the users asks for it and
doesn't write back anything that has not changed.

The `OverlayPreferenceStore` is relatively inconvenient to use because
one has to specify which keys should be cached by it. It would be
better if it would cache and write back everything. Furthermore, there
is no automatic read on the underlying preference store when the key is
not specified. This makes the creation of such an overlay store
cumbersome. Nevertheless, it is still better than the hand-rolled
solution that was used before.

This fixes both the `AutoEditsPreferencePage` and the
`SaveActionsPreferencePage`.

Fixes #1002487

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/ui/internal/preferences/AutoEditsPreferencePage.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/ui/internal/preferences/AutoEditsPreferencePage.scala
@@ -1,6 +1,7 @@
 package org.scalaide.ui.internal.preferences
 
 import org.eclipse.core.runtime.preferences.AbstractPreferenceInitializer
+import org.eclipse.jdt.internal.ui.preferences.OverlayPreferenceStore
 import org.eclipse.jface.layout.TableColumnLayout
 import org.eclipse.jface.preference.PreferencePage
 import org.eclipse.jface.viewers.CheckStateChangedEvent
@@ -30,16 +31,25 @@ import org.scalaide.util.eclipse.SWTUtils._
 /** This class is referenced through plugin.xml */
 class AutoEditsPreferencePage extends PreferencePage with IWorkbenchPreferencePage {
 
-  private val prefStore = IScalaPlugin().getPreferenceStore()
-
   private var descriptionArea: Text = _
   private var configurationArea: Text = _
   private var viewer: CheckboxTableViewer = _
 
   private val settings = AutoEdits.autoEditSettings.toArray
 
-  private var changes = Set[AutoEditSetting]()
-  private var configurations = Set[(AutoEditSetting, String)]()
+  private val prefStore = {
+    val ps = IScalaPlugin().getPreferenceStore
+    val keys = settings flatMap { s ⇒
+      import OverlayPreferenceStore._
+      Seq(
+        new OverlayKey(BOOLEAN, s.id),
+        new OverlayKey(STRING, s.configId)
+      )
+    }
+    val store = new OverlayPreferenceStore(ps, keys.toArray)
+    store.load()
+    store
+  }
 
   override def createContents(parent: Composite): Control = {
     val base = new Composite(parent, SWT.NONE)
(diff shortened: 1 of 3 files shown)


org.scala-ide.sdt.core/src/org/scalaide/ui/internal/preferences/AutoEditsPreferencePage.scala
```diff
@@ -30,16 +31,29 @@ import org.scalaide.util.eclipse.SWTUtils._
...
+    val ps = IScalaPlugin().getPreferenceStore
+    val keys = settings flatMap { s ⇒
+      import OverlayPreferenceStore._
+      Seq(
+        new OverlayKey(BOOLEAN, s.id),
+        new OverlayKey(STRING, s.configId)
+      )
+    }
+    val store = new OverlayPreferenceStore(ps, keys.toArray)
+    settings foreach { s ⇒
```
dragos: You could use `store.load()`. Have a look at `FormatterPreferencePage`, it's using the same technique

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/preferences/AutoEditsPreferencePage.scala
```diff
@@ -65,7 +79,7 @@ class AutoEditsPreferencePage extends PreferencePage with IWorkbenchPreferencePa
       }
     }
     viewer.addCheckStateListener { e: CheckStateChangedEvent =>
-      toggleAutoEdit(e.getElement().asInstanceOf[AutoEditSetting])
+      prefStore.setValue(e.getElement.asInstanceOf[AutoEditSetting].id, e.getChecked)
```
dragos: I don't see why you need to do this eagerly. Why not call `e.getChecked` on Apply or OK, for each checkbox?

dragos:
The `EditorPreferencePage` is using the same hand-rolled model, it would be good to move to this simpler way of doing things. Though, I am pretty sure even the `OverlayStore` is not needed.

Any reason why not using the `FieldEditorPreferencePage` in this cases? I think the code is simpler that way (have a look at `DebuggerPreferencePage`.

sschaef:
You are right, the `OverlayPreferenceStore` is not needed. Reading the values from the GUI elements is everything that is needed.

sschaef:
Actually, I no longer think we should go with the easy solution. The premature optimization that with the overlay store, no unnecessary writes happen to the real preference store is not important. But with the easy solution we limit ourselves to GUI elements that can store state for each preference. This doesn't work for example in the `AutoEditsPreferencePage`, where only one configuration text area exists, which is shared by all preferences. We would have to limit GUI design to our underlying possibilities. WDYT?

dragos:
What kind of additional configuration are there for auto-edits?

sschaef:
Right now there is no further configuration. It would be needed once the multi line string auto edit is moved to the new API. People could configure there if for example print margins should be added. Right now it needs to be configured in the editor page. We can just leave it there for now (I don't have time to move the last auto edit strategies to the new API anyway), remove the configuration option and reintroduce it the moment when we really need it.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/preferences/AutoEditsPreferencePage.scala
sschaef: /applied

sschaef:
I let the code as is and just added the `store.load()` function. I didn't change `EditorPreferencePage` because it works for now. Moving to `FieldEditorPreferencePage` would require more time and wouldn't really improve the situation, therefore I also skipped that. I also didn't remove the configuration option for auto edits.

dragos:
I can't try this right now, but if anyone confirms this works, I'm happy with the code :+1:

dragos:
LGTM

PR #971, opened 2015-06-18
Resolve execution conflict between surround selection and close pair

sschaef:
The problem was that when both surround selection and the close pair
auto edit were enabled, one of them is preferred over the other. To
solve this conflict some priority had to be introduced. Right now an
easy was is chosen - the close pair auto edits ensure that the surround
selection auto edits are not enabled.

Furthermore, `SurroundSelectionWithAngleBrackets` had to be introduced
to allow the corresponding close pair auto edit to check if it exists.

Fixes #1002488

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/core/internal/extensions/AutoEdits.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/core/internal/extensions/AutoEdits.scala
@@ -30,7 +30,8 @@ object AutoEdits {
     fullyQualifiedName[SurroundSelectionWithString] → SurroundSelectionWithStringSetting,
     fullyQualifiedName[SurroundSelectionWithParentheses] → SurroundSelectionWithParenthesesSetting,
     fullyQualifiedName[SurroundSelectionWithBraces] → SurroundSelectionWithBracesSetting,
-    fullyQualifiedName[SurroundSelectionWithBrackets] → SurroundSelectionWithBracketsSetting
+    fullyQualifiedName[SurroundSelectionWithBrackets] → SurroundSelectionWithBracketsSetting,
+    fullyQualifiedName[SurroundSelectionWithAngleBrackets] → SurroundSelectionWithAngleBracketsSetting
   )
 
   /**
--- a/org.scala-ide.sdt.core/src/org/scalaide/extensions/autoedits/CloseString.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/extensions/autoedits/CloseString.scala
@@ -3,6 +3,7 @@ package autoedits
 
 import org.eclipse.jdt.ui.text.IJavaPartitions
 import org.eclipse.jface.text.IDocument
+import org.scalaide.core.IScalaPlugin
 import org.scalaide.core.lexical.ScalaPartitions
 import org.scalaide.core.text.TextChange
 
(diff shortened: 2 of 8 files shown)


org.scala-ide.sdt.core/src/org/scalaide/extensions/autoedits/CloseString.scala
```diff
@@ -45,7 +46,9 @@ trait CloseString extends AutoEdit {
     document.textRangeOpt(offset-1, offset+1) exists (Set("{}", "[]", "()", "<>", "\"\"")(_))
 
   def autoClosingRequired(offset: Int): Boolean =
-    if (offset < document.length)
+    if (IScalaPlugin().getPreferenceStore.getBoolean(SurroundSelectionWithStringSetting.id))
```
dragos: What happens if I have both close parenthesis and surround selection enabled?

org.scala-ide.sdt.core/src/org/scalaide/core/internal/extensions/AutoEdits.scala
```diff
@@ -30,7 +30,8 @@ object AutoEdits {
     fullyQualifiedName[SurroundSelectionWithString] → SurroundSelectionWithStringSetting,
     fullyQualifiedName[SurroundSelectionWithParentheses] → SurroundSelectionWithParenthesesSetting,
     fullyQualifiedName[SurroundSelectionWithBraces] → SurroundSelectionWithBracesSetting,
-    fullyQualifiedName[SurroundSelectionWithBrackets] → SurroundSelectionWithBracketsSetting
+    fullyQualifiedName[SurroundSelectionWithBrackets] → SurroundSelectionWithBracketsSetting,
+    fullyQualifiedName[SurroundSelectionWithAngleBrackets] → SurroundSelectionWithAngleBracketsSetting
```
dragos: I would rather remove all angle-bracket auto-edits than add more (I understand it's because the logic requires it in the base class for all brackets.

Or, if you really want them for XML, can you add them only for XML partitions? I think auto-edits are usually enabled per partition type.

org.scala-ide.sdt.core/src/org/scalaide/extensions/autoedits/CloseString.scala
sschaef: In that case the surround selection is preferred.

org.scala-ide.sdt.core/src/org/scalaide/core/internal/extensions/AutoEdits.scala
sschaef: We could do that. On the other side it is already highly configurable. Users never need to enable them when they don't want them. That is a big plus to what we had before.

dragos: It's configurable per partition type. You can opt-in for having them everywhere or nowhere.. Maybe each auto-edit should have a set of (hard-coded) partitions where it's useful (so for the user it's still one on/off switch)

dragos:
I didn't realize this wasn't merged yet... I haven't tried it yet, but let's merge it :)

org.scala-ide.sdt.core/src/org/scalaide/core/internal/extensions/AutoEdits.scala
sschaef: Yes, the partitions are another problem domain. Don't know yet what is best to solve it. Everywhere vs nowhere sounds like a good solution when we don't want to force users to configure every partition.

PR #972, opened 2015-06-18
Remove stale HierarchyResolver.remember around advice.

stephenh:
The around code currently breaks running JUnit tests in Java projects that use
lambdas. This is because the cast to SourceTypeBinding fails on lambda bindings.
Eclipse suppresses the ClassCastException and tests just do not run.

As far as I can tell, there is actually no Scala IDE-specific code in this
"remember" advice anyway. It was added in 2009 by Miles, and looks exactly like
the code it's replacing in older releases of Eclipse:

https://github.com/eclipse/eclipse.jdt.core/blob/R3_2_maintenance/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/hierarchy/HierarchyResolver.java#L347

(E.g. unlike most advice, it was not calling proceed/etc., but instead was just
a raw copy/paste of the original code. Probably well intentioned while trying
to get things to work.)

However, newer versions of Eclipse have added an instanceof check to handle lambads:

https://github.com/eclipse/eclipse.jdt.core/blob/master/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/hierarchy/HierarchyResolver.java#L413

Which means this copy/paste of the code is now stale.

Instead of updating this code to match newer version of Eclipse, given
it looks unneeded, I've just removed it.

### DIFF
--- a/org.scala-ide.sdt.aspects/src-luna/scala/tools/eclipse/contribution/weaving/jdt/cfprovider/ClassFileProviderAspect.aj
+++ b/org.scala-ide.sdt.aspects/src-luna/scala/tools/eclipse/contribution/weaving/jdt/cfprovider/ClassFileProviderAspect.aj
@@ -111,11 +111,6 @@ public privileged aspect ClassFileProviderAspect {
     target(hr) &&
     args(suppliedType);
   
-  pointcut remember(HierarchyResolver hr, IType type, ReferenceBinding typeBinding) :
-    execution(void HierarchyResolver.remember(IType, ReferenceBinding)) &&
-    target(hr) &&
-    args(type, typeBinding);
-  
   pointcut acceptType(IType type, int acceptFlags, boolean isSourceType) :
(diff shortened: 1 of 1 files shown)


dragos:
ok to test

dragos:
LGTM. There is no reason to maintain that code anymore, you are absolutely right!

dragos:
Hi @stephenh,

Thankyou for your contribution.  Would you mind signing our CLA?

http://www.typesafe.com/contribute/cla

Thanks!

stephenh:
Sure, CLA signed.

dragos:
Excellent! as soon as the CI comes back with a build status I'll merge this. Thanks!

dragos:
Could you please open a ticket as well?

stephenh:
Yep, just did:

https://scala-ide-portfolio.assembla.com/spaces/scala-ide/tickets/1002489-cannot-run-junit-tests-that-use-java-8-lambdas#/activity/ticket:

Do I need to reference the ticket number in the commit message? Or is this good?

PR #974, opened 2015-06-28
Add Scala 2.12 Support

sschaef:
Beside from adding Scala 2.12 support this also fixes some other minor tickets.

See the commit messages for more detailed information about this PR.

### DIFF
--- a/org.scala-ide.sdt.core.tests/src/org/scalaide/core/classpath/ClasspathTests.scala
+++ b/org.scala-ide.sdt.core.tests/src/org/scalaide/core/classpath/ClasspathTests.scala
@@ -180,8 +180,6 @@ class ClasspathTests {
 
   /**
    * Std Library is the previous major version of Scala, with Xsource flag activated
-   *
-   * One warning witnessing a compatible version on classpath, which isnt exactly the one bundled.
    */
   @Test
   def previousLibraryWithXsource(): Unit = {
@@ -190,7 +188,7 @@ class ClasspathTests {
     projectStore.setValue(CompilerSettings.ADDITIONAL_PARAMS, "-Xsource:"+majorMinor)
     val newRawClasspath = cleanRawClasspath :+ createPreviousScalaLibraryEntry()
 
-    setRawClasspathAndCheckMarkers(newRawClasspath :+ newLibraryEntry("specs2_%s.2-0.12.3.jar".format(majorMinor)), expectedWarnings = 1, expectedErrors = 0)
+    setRawClasspathAndCheckMarkers(newRawClasspath :+ newLibraryEntry("specs2_%s.2-0.12.3.jar".format(majorMinor)), expectedWarnings = 0, expectedErrors = 0)
   }
 
   @Test
@@ -200,7 +198,7 @@ class ClasspathTests {
     projectStore.setValue(CompilerSettings.ADDITIONAL_PARAMS, s"-Xsource:$majorMinor")
     val newRawClasspath = cleanRawClasspath :+ createSubsequentScalaLibraryEntry()
 
-    setRawClasspathAndCheckMarkers(newRawClasspath :+ newLibraryEntry(s"specs2_$majorMinor.2-0.12.3.jar"), expectedWarnings = 1, expectedErrors = 0)
+    setRawClasspathAndCheckMarkers(newRawClasspath :+ newLibraryEntry(s"specs2_$majorMinor.2-0.12.3.jar"), expectedWarnings = 0, expectedErrors = 0)
   }
 
   @Test
@@ -224,7 +222,7 @@ class ClasspathTests {
   /**
    * Std Library is the previous major version of Scala, with Xsource flag activated, but binaries on classpath don't match
    *
-   * One warning witnessing a compatible version on classpath, one error on classpath validation
+   * One error on classpath validation
    */
   @Test
   def previousLibraryWithXsourceAndBadBinary(): Unit = {
@@ -233,7 +231,7 @@ class ClasspathTests {
     val majorMinor = getIncompatibleScalaVersion
     val newRawClasspath = cleanRawClasspath :+ createPreviousScalaLibraryEntry()
 
-    setRawClasspathAndCheckMarkers(newRawClasspath :+ newLibraryEntry("specs2_%s.2-0.12.3.jar".format(majorMinor)), expectedWarnings = 1, expectedErrors = 1)
+    setRawClasspathAndCheckMarkers(newRawClasspath :+ newLibraryEntry("specs2_%s.2-0.12.3.jar".format(majorMinor)), expectedWarnings = 0, expectedErrors = 1)
   }
 
   @Test
@@ -243,7 +241,7 @@ class ClasspathTests {
     val majorMinor = getIncompatibleScalaVersion
     val newRawClasspath = cleanRawClasspath :+ createSubsequentScalaLibraryEntry()
 
-    setRawClasspathAndCheckMarkers(newRawClasspath :+ newLibraryEntry(s"specs2_$majorMinor.2-0.12.3.jar"), expectedWarnings = 1, expectedErrors = 1)
+    setRawClasspathAndCheckMarkers(newRawClasspath :+ newLibraryEntry(s"specs2_$majorMinor.2-0.12.3.jar"), expectedWarnings = 0, expectedErrors = 1)
   }
 
   /**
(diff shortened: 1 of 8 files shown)


org.scala-ide.sdt.core/src/org/scalaide/core/internal/ScalaPlugin.scala
```diff
@@ -203,10 +201,11 @@ class ScalaPlugin extends IScalaPlugin with PluginLogConfigurator with IResource
             }
           }
           false
-        } else
-        true
+        }
+        else
```
dragos: nitpick, but please don't use this style. Prefer Java-style if-then-elses

org.scala-ide.sdt.core.tests/src/org/scalaide/core/classpath/ClasspathTests.scala
```diff
@@ -63,14 +64,14 @@ class ClasspathTests {
...
+  val baseRawClasspath = project.javaProject.getRawClasspath()
 
   /**
    * The classpath, with the eclipse scala container removed.
    */
-  private def cleanRawClasspath= for (classpathEntry <- baseRawClasspath
-        if classpathEntry.getPath().toPortableString() != "org.scala-ide.sdt.launching.SCALA_CONTAINER")
-      yield classpathEntry
+  private def cleanRawClasspath = for (
+    classpathEntry <- baseRawClasspath if classpathEntry.getPath().toPortableString() != "org.scala-ide.sdt.launching.SCALA_CONTAINER"
```
romanowski: this line could be break  or condition could be moved to def

org.scala-ide.sdt.core/src/org/scalaide/core/internal/ScalaPlugin.scala
```diff
@@ -53,29 +53,27 @@ object ScalaPlugin {
...
+  * 2.8.1 and 2.9.0 aren't.
+  */
+  def isCompatibleVersion(version: ScalaVersion, project: ScalaProject): Boolean = project.getCompatibilityMode match {
+    case Same ⇒
+      isBinarySame(ScalaVersion.current, version) // don't treat 2 unknown versions as equal
+    case Previous ⇒
       isBinaryPrevious(ScalaVersion.current, version)
-    else
-      isBinarySame(ScalaVersion.current, version)// don't treat 2 unknown versions as equal
+    case Subsequent ⇒
```
dragos: Would it be hard to make it work with any version, not just +/- 1? It seems like the same mechanism could be used to run, for instance, a build with Scala 2.9. Hypothetically, of course :)

org.scala-ide.sdt.core.tests/src/org/scalaide/core/classpath/ClasspathTests.scala
```diff
@@ -240,7 +297,8 @@ class ClasspathTests {
       expectedWarnings = 0, expectedErrors = 2)
   }
 
-  /** Major binary-incompatible library on the classpath, with and without version suffix of their own.
+  /**
```
dragos: We should converge on one style for comments. I am fine with both, so we can go for this one if you prefer it.

org.scala-ide.sdt.core/src/org/scalaide/core/internal/ScalaPlugin.scala
sschaef: That could be done, but I'm not sure if it is necessary to further subdivide the `Previous` and `Subsequent` part. 2.9 is as 2.10 previous and could therefore handled by the same logic. I thought about potential dotty support and it could be added by just extending the trait `CompatiblityMode`.

org.scala-ide.sdt.core.tests/src/org/scalaide/core/classpath/ClasspathTests.scala
sschaef: We did in some other PR. The result was that we want to use

```
/**
 * Text here
 */
```

for multi line Scaladoc and `/** Text here */` for single line Scaladoc. There is just not yet a style guide somewhere that mentions this.

dragos: a, ok. I forgot :)

org.scala-ide.sdt.core/src/org/scalaide/core/internal/ScalaPlugin.scala
dragos: So basically there's only 3 cases:
- same version
- older version
- newer version

but not necessarily only 1 major version before/after?

sschaef: That is correct, would you like to see a change here then? `Subsequent` can be renamed to `ScalaVersion212` and `Previous` to `ScalaVersion210`.

dragos: I think that's the opposite of what I was asking. Previous means any previous version, 2.10, 2.9, 2.8? and Subsequent means 2.12, 2.13, etc? I think it should work the same way for any version, right?

sschaef: Not sure if I understand what you mean. Technically, both previous and subsequent versions can be treated in the same way but I don't think that simplifies anything. We need to be able to detect when a user switches from say 2.10 to 2.12.

dragos: I guess I could just try it, but my question was: does this PR allow me to create a Scala installation for Scala 2.8.0 and use it to build a project? Or it limits the user to 2.10, 2.11 and 2.12?

sschaef: It should limit them to 2.10+ since the `-Xsource` flag is set, which doesn't exist for <2.10.

org.scala-ide.sdt.core.tests/src/org/scalaide/core/classpath/ClasspathTests.scala
```diff
@@ -524,13 +584,13 @@ class ClasspathTests {
     // two excepted code errors
     var markers = project.underlying.findMarkers(SdtConstants.ProblemMarkerId, true, IResource.DEPTH_INFINITE)
     val errors = SDTTestUtils.markersMessages(markers.toList)
-    assertEquals("Unexpected number of scala problems in project: " + errors, 3, errors.length)
+    assertEquals("Unexpected number of scala problems in project: "+errors, 3, errors.length)
```
dragos: Space around `+`

org.scala-ide.sdt.core/src/org/scalaide/core/internal/project/ClasspathManagement.scala
```diff
@@ -371,57 +371,61 @@ trait ClasspathManagement extends HasLogger { self: ScalaProject =>
...
+          ClasspathErrorMarker(IMarker.SEVERITY_WARNING, msg, SdtConstants.ScalaVersionProblemMarkerId) :: Nil
         }
-      case _ => // 2 or more of them, not great, but warn only if the library is not a project
-        if (fragmentRoots.exists(incompatibleScalaLibrary))
-          ClasspathErrorMarker(IMarker.SEVERITY_ERROR, moreThanOneLibraryError(fragmentRoots.map(_.location), compatible = false), SdtConstants.ClasspathProblemMarkerId) :: Nil
+        // Previous version, and the XSource flag isn't there already : warn and suggest fix using Xsource
+        else ClasspathErrorMarker(IMarker.SEVERITY_ERROR, msg, SdtConstants.ScalaVersionProblemMarkerId) :: Nil
+      case Some(v) =>
+        // incompatible version
+        ClasspathErrorMarker(IMarker.SEVERITY_ERROR, s"The version of scala library found in the build path (${v.unparse}) is incompatible with the one expected by scala IDE ($expectedVersion). Please replace the scala library with the scala container or a compatible scala library jar.", SdtConstants.ClasspathProblemMarkerId) :: Nil
```
dragos: Shouldn't this allow a `Subsequent` libraries? If there's only one scala lib, 2.12, will it work?

dragos:
I think this got out of sync with `scala-refactoring` and I can't easily build in my IDE. It'd be great to rebase so I can give it a try.

sschaef:
rebased to master, no further changes (regarding the nitpicks) included

PR #977, opened 2015-07-10
Jdi sync with java compiling fixes #1002486

wpopielarski:
it fixes sbt compilation problem when java class depends on scala class. Quite descriptive test of scenario can be found in JavaDependsOnScalaTest.scala.


wpopielarski:
covered with other PRs

PR #979, opened 2015-07-17
Disable debug tracing in move class refactoring

mlangc:
Instantiating refactorings with tracing enabled is not recommended for production. Apart from that `ConsoleTracing` is about to be renamed in the refactoring library.


mlangc:
Here is the associated PR to scala-refactoring: https://github.com/scala-ide/scala-refactoring/pull/87

sschaef:
LGTM

PR #987, opened 2015-07-26
Skip expensive buildStructure calls when `NO_AST` and no `resolveBind…

dragos:
…ings` are needed.

This code path is taken by `Save`, on the UI thread, and
slow build structure can block the UI for several seconds while saving.

In order to not lose functionality, the structure is now built after each reconciliation. It should be fast (the whole file was already type-checked), and on a background thread.

These changes interfere with the JDT interactions so it'd be great to get some feedback from real-world usage.

Fixed #1002412

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/core/internal/jdt/model/ScalaSourceFile.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/core/internal/jdt/model/ScalaSourceFile.scala
@@ -115,8 +115,11 @@ class ScalaSourceFile(fragment : PackageFragment, elementName: String, workingCo
 
     // don't rerun this expensive operation unless necessary
     if (!isConsistent()) {
-      val info = createElementInfo.asInstanceOf[OpenableElementInfo]
-      openWhenClosed(info, true, monitor)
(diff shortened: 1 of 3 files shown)


org.scala-ide.sdt.core/src/org/scalaide/core/internal/jdt/model/ScalaSourceFile.scala
```diff
@@ -103,7 +103,7 @@ class ScalaSourceFile(fragment : PackageFragment, elementName: String, workingCo
       monitor : IProgressMonitor) : org.eclipse.jdt.core.dom.CompilationUnit = {
     /* This explicit call to super matters, presumably exercised
       through AspectJ. See #1002016. */
```
sschaef: This comment should be removed too.

sschaef:
I like this PR but first have to try it out in order to find out if it works reliable.

dragos:
@sschaef did you get the chance to try it in practice? I'm using it for the past few days, and it works well for me.

sschaef:
Yes, I used it the last days and it worked fine.

sschaef:
LGTM

PR #988, opened 2015-07-29
Fix #1002518

dragos:
Reorder auto-edits to give priority to more specific actions. Close parentheis/braces/etc is the most generic, fallback type of action.


sschaef:
LGTM

PR #989, opened 2015-07-29
A few fixes to surround block auto-edit.

dragos:
- stop at comment blocks
- don’t include empty lines in the surround block
- properly treat `else` 

Fix #1002515
Fix #1002519

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/extensions/autoedits/SurroundBlock.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/extensions/autoedits/SurroundBlock.scala
@@ -45,6 +45,7 @@ trait SurroundBlock extends AutoEdit {
   override def setting = SurroundBlockSetting
 
   override def perform() = {
+    val elseLikeTokens = Set(Tokens.ELSE, Tokens.CATCH, Tokens.FINALLY)
     check(textChange) {
       case Add(start, "{") =>
         surroundLocation(start) map {
@@ -52,7 +53,7 @@ trait SurroundBlock extends AutoEdit {
             val sep = System.getProperty("line.separator")
             val indent = " " * indentLen
 
-            val change = if (token.tokenType == Tokens.ELSE)
+            val change = if (elseLikeTokens(token.tokenType))
               Replace(start, pos + indentLen, s"{${document.textRange(start, pos)}$indent} ")
             else
               Replace(start, pos, s"{${document.textRange(start, pos)}$indent}$sep")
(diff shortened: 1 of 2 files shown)


org.scala-ide.sdt.core/src/org/scalaide/extensions/autoedits/SurroundBlock.scala
```diff
@@ -92,10 +102,8 @@ trait SurroundBlock extends AutoEdit {
 
             if (prevLine.start == firstLine.start)
               None
-            else if (t.tokenType == Tokens.VARID)
```
dragos: This part I didn't understand what it was about. After removing it, no test failed, and it's what helped fix my bug about including blank lines in surround block.

dragos:
@sschaef it'd be good if you had a look, since this is your work.

org.scala-ide.sdt.core/src/org/scalaide/extensions/autoedits/SurroundBlock.scala
```diff
@@ -63,7 +68,7 @@ trait SurroundBlock extends AutoEdit {
    *
    * In case no insertion position could be found, `None` is returned.
    */
-  private def surroundLocation(offset: Int): Option[(Int, Int)] = {
+  private def surroundLocation(offset: Int): Option[(Int, Int, Token)] = {
```
sschaef: Scaladoc should be adapted to the new return type.

org.scala-ide.sdt.core/src/org/scalaide/extensions/autoedits/SurroundBlock.scala
```diff
@@ -48,10 +48,15 @@ trait SurroundBlock extends AutoEdit {
...
       case Add(start, "{") =>
         surroundLocation(start) map {
-          case (pos, indentLen) =>
+          case (pos, indentLen, token) =>
             val sep = System.getProperty("line.separator")
-            Replace(start, pos, "{" + document.textRange(start, pos) + " "*indentLen + "}" + sep)
-              .withCursorPos(start+1)
+            val indent = " " * indentLen
+
+            val change = if (token.tokenType == Tokens.ELSE)
```
sschaef: We could also check for a `catch` here.

org.scala-ide.sdt.core/src/org/scalaide/extensions/autoedits/SurroundBlock.scala
sschaef: I can't remember for what that case was needed. It is very well possible that it is no longer needed.

org.scala-ide.sdt.core/src/org/scalaide/extensions/autoedits/SurroundBlock.scala
sschaef: A `case` block is another use case.

dragos: Yeap, catch/finally/yield.

dragos: Actually, `yield` doesn't make sense.

sschaef:
LGTM

PR #990, opened 2015-07-30
Also treat block comments/scaladoc as comments

dragos:
Fix #1002515 (again).

@sschaef I missed a case...


sschaef:
`COMMENTS` is a `Set`, I wonder why the compiler didn't emit a warning that the `==` always will yield false.

sschaef:
LGTM

PR #991, opened 2015-08-05
Async debugger

sschaef:
This is not yet completely finished work, the following still needs to be done:

- [x] Clear Async Stack view when program is terminated
- [x] Disable bang functionality, because it doesn't work correctly yet
- [x] Make filter in async stack configurable
- [ ] Huge slowdown of the async debugger (it looks like it is even in some endless waiting state) when the async stack view is opened before the debugger is started.
- [ ] Enable/disable async stack collection per Run Configuration. It is possible to disable it globally through preferences but that is not the best solution.

Delayed for at least 4.2RC2:

- [ ] Add functionality to jump back to start when in Async Stack
- [ ] Key combination for the "jump to next message" feature
- [ ] The "jump to next message" icon should be shown in the debug view and not in the main menu bar
- [ ] Dead letter functionality is not working. This is because right now no BreakpointEvents are sent to the corresponding component.

PR #992, opened 2015-08-12
Async debugger actor

wpopielarski:
It implements very basic functionality of async debugger.
The purpose of async debugger is to ease debugging of applications
where objects exchange messages rather than call synchroniously
methods. Right now it is possible to track messages exchanged by Akka
actors, old scala actors and futures.

- [x] check if looped actor message handling is correctly shown in Async Stack
- [ ] some tests to cover basic use cases


dragos:
What's the relation to #991 ? Does one supersede the other? Please close the irrelevant one :)

wpopielarski:
aik #991 is to close but S. needs to roger that. I'm working on some nice
workaround for object message now.

2015-08-13 12:00 GMT+02:00 Iulian Dragos <notifications@github.com>:

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/model/VirtualValue.scala
```diff
@@ -0,0 +1,47 @@
...
+  override def getReferenceTypeName(): String = refTypeName
+
+  override def getValueString(): String = valueString
+
+  def withFields(vars: IVariable*) =
+    VirtualValue(refTypeName, valueString, fields ++ vars.toSeq)
+
+  override def getVariables(): Array[IVariable] = fields.toArray
+
+  override def hasVariables(): Boolean = fields.isEmpty
```
dragos: I think this would normally be `!fields.isEmpty`, right? (my mistake originally)

wpopielarski: `fields.nonEmpty` but I'm wonder what impact of it is on functionality?

wpopielarski: ok I see it, it's a part of logical structure, I didn't click on it too much.

dragos: I don't know, but seems wrong as it is, so let's fix it :)

wpopielarski: done

2015-08-14 15:06 GMT+02:00 Iulian Dragos <notifications@github.com>:

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/async/RetainedStackManager.scala
```diff
@@ -0,0 +1,135 @@
...
+    case k @ (objRef, _) if objRef == future => k
+  }
+
+  private def keysSelection(future: ObjectReference): PartialFunction[Int, PartialFunction[AsyncStackTraceKey, AsyncStackTraceKey]] = {
+    case ordinal if ordinal != OrdinalNotSet => keysForObjectReferenceAndOrdinal(future, ordinal)
+    case OrdinalNotSet => keysForObjectReference(future)
+  }
+
+  private def previousKeyWhenIsGiven(future: ObjectReference, ordinal: Int): AsyncStackTraceKey = {
+    val olderTracesKeys = stackFrames.keys.collect(keysSelection(future)(ordinal))
```
dragos: I don't like that, for the special-case of objects, we start doing a linear search for everybody, instead of a hashmap lookup. There should be a fast-path for objects that are unique. If there's only one entry for some `future` (we also need to rename this, because it's not a future, it can be a message as well), there's no need to look further. So maybe go back to a `Map[ObjectReference, <something else here>]`, and have a fast-path when there's only one of those, and otherwise use the `ordinal` to disambiguate.

Also, please add an explanation in the scaladoc comment of `getStackFrameForFuture` (also, let's rename it to reflect that it's not only for futures... maybe simply `getAsyncStackFrame`).

wpopielarski: got is and will ponder it. generally if something is unique or not we know
in moment of insertion and than we can decide how to handle it.

2015-08-14 16:18 GMT+02:00 Iulian Dragos <notifications@github.com>:

wpopielarski: but I see now concept is clear :)

2015-08-14 16:46 GMT+02:00 Wieslaw Popielarski <wpopielarski@gmail.com>:

wpopielarski: just lightning thought: [ObjectReference, List[<something>]] is a good
trace?

2015-08-14 16:46 GMT+02:00 Wieslaw Popielarski <wpopielarski@gmail.com>:

wpopielarski: you right, I try something like Map[ObjectReference, AsyncStackTraces]
where AsyncStackTraces(listOfOrdinals: List[Int], asyncStackTracesMap[Int,
AsyncStackTrace])
usage:

val asyncStackTrace = map(objRef)(ordinal)

class AsyncStackTraces(listOfOrdinals: List[Int], asyncStackTracesMap[Int,
AsyncStackTrace]) {
  def apply(index: Int) = {
if (listOfOrdinals.tail == Nil) asyncStackTraces(listOfOrdinals.head) else
asyncStackTraces(findPreviousIndex(listOfOrdinals, index)) }

this should be fine I guess. It's almost constant searching for map and log
n  for indices on list. Is it ok?

2015-08-14 16:58 GMT+02:00 Wieslaw Popielarski <wpopielarski@gmail.com>:

dragos: I think `Map[ObjectReference, List[somthing]]` is a good one. I think linear search through the second list is ok.

wpopielarski: ok

2015-08-17 10:31 GMT+02:00 Iulian Dragos <notifications@github.com>:

dragos:
By the way, I just realized. Not part of this PR, but we should do it before the final release:

- a way to enable/disable async stack collection. This should be a setting per **Run Configuration**, if possible. This way people can run the debugger with or without async features (if it proves too much overhead).

sschaef:
That is a good idea, I just have no idea how to add such a switch to the run configuration. Adding a switch in the preferences is easier but would not work for single run configurations.

wpopielarski:
roger that. I can look at it.


2015-08-17 12:47 GMT+02:00 Iulian Dragos <notifications@github.com>:

sschaef:
This PR is superseded by #991 

I merged the last two commits to the mentioned PR, but not the `after merge` commit since it is a weird merge commit. It somehow duplicates the content, I don't know what you did there.

wpopielarski:
ok, if you merged changes from 992 to 991. 992 was created to cut off
'jumped stops' and 'dead letter stop'.

2015-08-17 18:46 GMT+02:00 Simon Schäfer <notifications@github.com>:

PR #994, opened 2015-08-24
Async debugger with bang operator

wpopielarski:
Enhancement added to Async Stack Trace functionality
- The tab has been added to 'Launch Configuration' to control if async debugger garners or not async stack traces.
- bang operator (message step out) and stop on dead letter functionalities has been activated


org.scala-ide.sdt.core/src/org/scalaide/core/internal/async/AsyncDebuggerTab.scala
```diff
@@ -0,0 +1,123 @@
...
+import org.scalaide.ui.ScalaImages
+
+import AsyncDebuggerConfiguration.LaunchWithAsyncDebugger
+
+private class GeneralSettingsBlock(val tab: AsyncDebuggerTab) {
+  import AsyncDebuggerConfiguration._
+
+  private var globalAsyncDebuggerSwitchButton: Button = _
+  val SettingsArea = "General Settings"
+  val AsyncDebuggerGlobalSwitch = "Switch On/Off Async Debugger functionality for this debug configuration"
```
dragos: I think "on off" is redundant when there's a check box right next to it. How about "Enable Async Debugger functionality".

dragos:
Please update the title of this PR to reflect what it is about

dragos:
We need to make this available to other launch configurations, I think:

- add a Scala JUnit launcher, similar to the Java one, with this additional tab group
- corresponding change in the ScalaTest plugin
- check that a Java Akka application can still use the async debugger.. I wonder if a Java application can be launched using the Scala Application launcher (then we don't need to do anything here, just document it)
- Eclipse application (at least the one with weaving), so we can also debug (at least the worksheet is using Akka).

Given how many configurations we need, I wonder if it's better to use some flag, after all.

org.scala-ide.sdt.core/src/org/scalaide/core/internal/async/AsyncDebuggerTab.scala
```diff
@@ -0,0 +1,123 @@
...
+      configuration.removeAttribute(LaunchWithAsyncDebugger)
+    }
+
+  /**
+   * @return a string error message or <code>null</code> if the block contents are valid
+   */
+  def validate: String = null
+}
+
+class AsyncDebuggerTab extends AbstractLaunchConfigurationTab {
```
sschaef: If we add a tab to the launch configuration, I would make it more general. A "Scala Debugger" tab that contains sections about different functionality is better for future additions.

wpopielarski: makes sense

wpopielarski:
@dragos 
> - add a Scala JUnit launcher, similar to the Java one, with this additional tab group

Last commit covers it
> - corresponding change in the ScalaTest plugin

I'm going to open PR, change is quite simple
> - check that a Java Akka application can still use the async debugger.. I wonder if a Java application can be launched using the Scala Application launcher (then we don't need to do anything here, just document it)

Generally Async Tracer works with Akka Java API too. The problem is that Scala Launcher doesn't recognizes the Java main as executable and I think we should if project got Scala Nature. But this is not related to Async feature.
> - Eclipse application (at least the one with weaving), so we can also debug (at least the worksheet is using Akka).

ok, is coming

> Given how many configurations we need, I wonder if it's better to use some flag, after all.

well it was my first proposal :)

org.scala-ide.sdt.core/src/org/scalaide/core/internal/async/AsyncDebuggerTab.scala
wpopielarski: roger

dragos:
So, for a Java Akka application, after these changes, it won't be possible to use the Async Debugger, right? Since the Scala Launcher won't recognize it, and the Java one doesn't have the checkbox (and the default is `off`).

wpopielarski:
Hey good men
we have a verifier of main class (`MainClassVerifier`) which wants from main class to be an scala object. I just commented out this piece of code
<code>
  private def canRunMain(project: IScalaProject, mainTypeName: String, hasBuildErrors: Boolean): IStatus = {
    val mainClass = findClassFile(project, mainTypeName)
    def mainModuleClass = findClassFile(project, mainTypeName)// + MainClassVerifier.ModuleClassSuffix) <--- HERE IS MY CHANGE
</code>
and it is working. Do we want it permanently?

dragos:
Just removing it will probably fail a bunch of tests. The restriction is there so that people don't make the mistake of writing a Scala class with a `def main`, but that won't work because it's not static. @dotta worked on this, IIRC.

We need to be smarter about this check, probably by loading the class reflectively and checking if there's a static main method in there.

dragos:
LGTM, but I think you should add the tab for "JUnit plugin test" as well..

wpopielarski:
Ok, I'll add it. Meantime 'main' is solved I hope

rkrzewski:
What about launchers provided by 3rd party plugins? I'm using http://bndtools.org/ together with ScalaIDE for one of my hobby projects. It works quite well, but I wasn't able to use any of the new Scala debugger features, because Bndtools provide their own launcher for starting OSGi frameworks. 
Can this be solved generically on ScalaIDE side, or an separate integration plugin would be needed for any tool that provides it's own launchers?

wpopielarski:
@rkrzewski 
not sure about it. I just see that you can define launch delegate and here is no extension point for debugger or saying more general for VM runner. So probably we would need to implement Scala launch delegate with such extension point. Now we deliver the `org.scalaide.debug.internal.launching.ScalaDebuggerForLaunchDelegate` trait you can mix with your delegate.

rkrzewski:
@wpopielarski mix with my what?
I take that to mean it won't work, unless somebody writes an integration plugin :disappointed:

wpopielarski:
@rkrzewski
mix with your launch delegate implementation

2015-08-26 14:12 GMT+02:00 Rafał Krzewski <notifications@github.com>:

wpopielarski:
@rkrzewski
ok I got it, you are just using 2 3-rd parties: bndtools and scala ide
(eclipse plugins). And bndtools has its own launcher. And you want to run
scala debugger in bndtools launcher. Well, I don't see the way to do this
now. Eclipse launcher doesn't expose an extension point to plug a runner.
You need to create your own eclipse plugin.

2015-08-26 14:24 GMT+02:00 Wieslaw Popielarski <wpopielarski@gmail.com>:

dragos:
@rkrzewski indeed, a plugin needs to provide another launch delegate for the BndTools launch configuration. The Scala IDE plugin does it for a number of existing launch configurations, you can see the code [here](https://github.com/scala-ide/scala-ide/blob/master/org.scala-ide.sdt.debug/plugin.xml#L26). You can try adding it there, and if it works just by an ID, and does not require BndTools as a dependency, open a PR and we'll add it.

wpopielarski:
Scala Debugger Tab for JUnit Plug-in Test added

rkrzewski:
@dragos I don't see how it could be done without dependency on Bndtools code, for example [ScalaJUnitLaunchConfigurationDelegate](https://github.com/scala-ide/scala-ide/blob/master/org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/launching/ScalaJUnitLaunchConfigurationDelegate.scala) extends `org.eclipse.jdt.junit.launcher.JUnitLaunchConfigurationDelegate`

I'm not sure what you mean by "working just by ID" here. The last time I dabbled in Eclipse plugin development was in 2009, so please bear with me :)

I also find https://github.com/scala-ide/scala-ide/blob/master/org.scala-ide.sdt.debug/plugin.xml#L47 declaration rather confusing, is the `name` attribute wrong, perhaps?

wpopielarski:
@rkrzewski
not sure but seems that @dragos means it's enough to define id (required by
schema) and type which points to launch configuration type (defined in 3-rd
party plugin.xml) and then the delegate is taken from that configuration

2015-08-26 15:01 GMT+02:00 Rafał Krzewski <notifications@github.com>:

sschaef:
Another important configuration in my opinion is "Remote Java Application". Would you please also add the tab to this run configuration?

sschaef:
Wieslaw, please don't add commits to a branch which should not be part of the associated PR (as it is the case for the last two commits.)

wpopielarski:
I don't want to have two PRs. I think it should come with one. This is
still the same Async Debugger and in final release will be complete.

2015-08-27 15:57 GMT+02:00 Simon Schäfer <notifications@github.com>:

sschaef:
On 08/27/2015 05:19 PM, wpopielarski wrote:
> I don't want to have two PRs. I think it should come with one. This is
> still the same Async Debugger and in final release will be complete.
Well, no. The tab page is a separate feature. It doesn't have anything 
to do with the other async debugging improvements that still need to be 
done. If you keep the PRs small, we can merge more often and therefore 
the risk for rebasing with merge errors (and your need to stay on one 
branch) goes down.

wpopielarski:
Let's talk tomorrow about that. I had 3 reasons to do is in this way.
Anyway I'm not sure if we need to break lances about it.

2015-08-27 17:36 GMT+02:00 Simon Schäfer <notifications@github.com>:

sschaef:
LGTM

PR #995, opened 2015-08-27
Improve async debugger

sschaef:
- Re-enables the disabled "step to next message" feature
- Adds preference page to allow users to configure async program points
- Allow debugger commands to have key combinations (no default key combinations set, since it is difficult to find one that works on all systems)
- Move "step to next message" icon to async view toolbar


dragos:
It seems both #994 and this PR added back "bang"... so which one is the right one? :)

sschaef:
I'm going to merge the other PR first and then rebase this one on top of master. Then, the first commit should be gone.

dragos:
Looks good, I like the new configuration dialog!

sschaef:
Superseded by #997

PR #996, opened 2015-08-30
Remove source folder src-kepler and integrate src-luna in src

sschaef:
We no longer support Kepler, therefore we don't need this source folder
anymore. Since Luna is the only supported platform right now and Mars
seems to work fine with our current codebase, there is no need to keep
src-lune. Therefore its content is integrated in src.

Fixes #1002530


dragos:
I'm not sure our codebase builds in Mars. I think someone reported some source incompatibilities (but the binaries work fine in Mars).

sschaef:
Ok, I'm going to find out.

sschaef:
It doesn't compiler under Luna :( I'm going to fix that.

PR #998, opened 2015-08-31
Remove dependency to VMHelper

sschaef:
This class is part of a package in PDE, which we do not want to rely on
since it is only available when PDE is installed in Eclipse.

Fixes #1002538


sschaef:
Review @dragos and @wpopielarski. But I don't think there is much to review. I tested it and it works fine for both ScalaTest and the Equinox Weaving launcher. I must admit, I don't know what the difference to the `VMHelper` class is but as long as it works I don't care that much.

wpopielarski:
hehehe, you are like this merchant of pills to quench thirst in the Little
Prince. So you just saved 53 minuts instead of going to the source :). but
I understand

2015-08-31 20:02 GMT+02:00 Simon Schäfer <notifications@github.com>:

PR #999, opened 2015-09-02
Build Scopes Configuration

wpopielarski:
There are 3 build scopes defined: macros, main and tests. Up to now
there was a convention how source folders were assigned to particular
scope. This commit adds a configuration which allows Scala IDE user to
assign the particular source folder to required scope via 'Scopes
Settings' tab in 'Scala Compiler' preference page.
Known limitation: when project's 'Properties' is open and user adds
new source folder in 'Java Build Path -> Source' the 'Properties'
popup window must be reopened to make visible newly added source path
in 'Scala Compiler -> Scope Settings'.

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/ui/internal/preferences/ScopesSettings.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/ui/internal/preferences/ScopesSettings.scala
@@ -1,0 +1,36 @@
+package org.scalaide.ui.internal.preferences
+
+import scala.tools.nsc.Settings
+
+import org.eclipse.core.resources.IProject
+import org.eclipse.core.runtime.IPath
+import org.scalaide.core.internal.ScalaPlugin
+import org.scalaide.core.internal.project.CompileScope
+import org.scalaide.ui.internal.preferences.IDESettings.Box
+
+object ScopesSettings extends Settings {
+  val TabTitle = "Scopes Settings"
+  private val NamePrefix = "-"
+  private val EmptyString = ""
+
+  private def choices = CompileScope.scopesInCompileOrder.map { _.name }.toList
+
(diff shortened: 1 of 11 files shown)


org.scala-ide.sdt.core/src/org/scalaide/ui/internal/preferences/ScopesSettings.scala
```diff
@@ -0,0 +1,36 @@
...
+        ChoiceSetting(srcName, helpArg = EmptyString, descr = EmptyString, choices, findDefaultScope(srcFolderRelativeToProject))
+      }
+    }.getOrElse(Nil).toList.sortBy { _.name }
+  }
+
+  def makeKey(srcFolderRelativeToProject: IPath): String =
+    NamePrefix + srcFolderRelativeToProject.segments.mkString("/")
+
+  private def findDefaultScope(srcFolderRelativeToProject: IPath): String =
+    CompileScope.scopesInCompileOrder.find { _.isValidSourcePath(srcFolderRelativeToProject) }.get.name
```
sschaef: Is that `get` safe here?

sschaef:
Source code looks code but again, it would be nice if changes regarding formatting would be moved to their own commit.

wpopielarski:
but this is just autoformatter

2015-09-02 17:10 GMT+02:00 Simon Schäfer <notifications@github.com>:

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/preferences/ScopesSettings.scala
wpopielarski: must not be the case when none of scopes is chosen

sschaef:
LGTM

PR #1000, opened 2015-09-03
Bang operator not working

wpopielarski:
There are two things fixed in this commit:
- command ids didn't reflect the id in command definition
- command must be associated with debug context
To explain the second it is enough to look at `StepMessageOut` class.
The first element of selection must be `ScalaStackFrame` and continue
action should run debugging further. In buggy solution the first element
of selection is `AsyncStackFrame` and action does not continue debugging.

### DIFF
--- a/org.scala-ide.sdt.debug/plugin.xml
+++ b/org.scala-ide.sdt.debug/plugin.xml
@@ -142,17 +142,20 @@
          point="org.eclipse.ui.menus">
       <menuContribution
             allPopups="false"
-            locationURI="toolbar:org.scala-ide.sdt.debug.asyncView">
-         <command
-               commandId="org.scala-ide.sdt.debug.stepMessageOut"
-               icon="icons/bang.png"
-               label="Step Message Out"
-               style="push"
-               tooltip="Step until a message is received">
-            <visibleWhen
-                  checkEnabled="true">
-            </visibleWhen>
-         </command>
+            locationURI="toolbar:org.eclipse.ui.main.toolbar">
+         <toolbar
+               id="org.eclipse.debug.ui.debugActionSet">
+            <command
+                  commandId="org.scalaide.debug.async.stepMessageOut"
+                  icon="icons/bang.png"
+                  label="Step Message Out"
+                  style="push"
+                  tooltip="Step until a message is received">
+               <visibleWhen
+                     checkEnabled="true">
+               </visibleWhen>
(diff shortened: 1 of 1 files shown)


dragos:
LGTM

org.scala-ide.sdt.debug/plugin.xml
```diff
@@ -142,17 +142,20 @@
...
-               commandId="org.scala-ide.sdt.debug.stepMessageOut"
-               icon="icons/bang.png"
-               label="Step Message Out"
-               style="push"
-               tooltip="Step until a message is received">
-            <visibleWhen
-                  checkEnabled="true">
-            </visibleWhen>
-         </command>
+            locationURI="toolbar:org.eclipse.ui.main.toolbar">
```
sschaef: Changing the locationURI is not correct. We want to have the icon in the Async Stack view (or in the Debug view) but not in th emain menu toolbar.

wpopielarski: I try to find it in eclipse doc. As I said it cannot be in Async View which
shows history not current state

2015-09-03 12:22 GMT+02:00 Simon Schäfer <notifications@github.com>:

wpopielarski: looks this is better but needs to be verified:
            locationURI="toolbar:org.eclipse.debug.ui.DebugView"

sschaef:
I fail to see what the location of the icon has to do with the underlying logic. If the underlying logic is broken it needs to be fixed.

sschaef:
I'm sure if would be possible to somehow get the correct selection, but the current solution works fine.

LGTM

wpopielarski:
well as I got it, you get selection from view you add your handler to its toolbar. I didn't delve into it.

dragos:
Hm, I didn't try this locally, but the bang button should be on the same toolbar as the other "step" buttons, right? Conceptually is very similar. View-specific buttons should do something to that view (filter, show/hide elements, etc.)

sschaef:
On 09/07/2015 02:54 PM, Iulian Dragos wrote:
>
> Hm, I didn't try this locally, but the bang button should be on the 
> same toolbar as the other "step" buttons, right? Conceptually is very 
> similar. View-specific buttons should do something to that view 
> (filter, show/hide elements, etc.)
>
Yes, this is how it is now implemented.

PR #1002, opened 2015-09-05
fix tooltip message for 'var's. (#1001007)

andrey-ilinykh:


sschaef:
ok to test

sschaef:
The build failed because you introduced trailing whitespace: https://jenkins.scala-ide.org:8496/jenkins/job/ghprb-scala-ide-multi-conf/eclipse_platform=eclipse-luna/316/console

Also, please use `if (...)` format-style for the ifs.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/hover/ScalaHoverImpl.scala
```diff
@@ -116,7 +116,21 @@ class ScalaHoverImpl extends ITextHover with ITextHoverExtension with ITextHover
...
+              Option("var " + sym.localName + ":" +gsym.tpe.resultType)
+            else
+              compiler.headerForSymbol(sym, tpe)
+          }
+          else  
+            compiler.headerForSymbol(sym, tpe)
+          
+            
+          
+        }).flatten
```
sschaef: This produces a race condition. Symbols need to be accessed on the compiler thread. Wrap the yield block in:
```scala
compiler.asyncExec(...).getOrElse(None)()
```

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/hover/ScalaHoverImpl.scala
```diff
@@ -116,7 +116,21 @@ class ScalaHoverImpl extends ITextHover with ITextHoverExtension with ITextHover
 
         val scalaRegion = new Region(icu.sourceMap(viewer.getDocument.get.toCharArray).scalaPos(region.getOffset), region.getLength)
 
-        def typeInfo(t: Tree): Option[String] = (for (sym <- Option(t.symbol); tpe <- Option(t.tpe)) yield compiler.headerForSymbol(sym, tpe)).flatten
+        def typeInfo(t: Tree): Option[String] = (for (sym <- Option(t.symbol); tpe <- Option(t.tpe)) yield {
+          if( sym.isMethod && sym.isSourceMethod ){//this is info message for 'var's. We check if owner has both setter and getter
+            val gsym = sym.getterIn(sym.owner)
+            val ssym = sym.setterIn(sym.owner)
+            if( gsym != NoSymbol && ssym != NoSymbol)
```
sschaef: This nested if can be removed by converting `gsym` and `ssym` to a def and then call them in the first if.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/hover/ScalaHoverImpl.scala
```diff
@@ -116,7 +116,21 @@ class ScalaHoverImpl extends ITextHover with ITextHoverExtension with ITextHover
 
         val scalaRegion = new Region(icu.sourceMap(viewer.getDocument.get.toCharArray).scalaPos(region.getOffset), region.getLength)
 
-        def typeInfo(t: Tree): Option[String] = (for (sym <- Option(t.symbol); tpe <- Option(t.tpe)) yield compiler.headerForSymbol(sym, tpe)).flatten
+        def typeInfo(t: Tree): Option[String] = (for (sym <- Option(t.symbol); tpe <- Option(t.tpe)) yield {
+          if( sym.isMethod && sym.isSourceMethod ){//this is info message for 'var's. We check if owner has both setter and getter
```
sschaef: This comment is unnecessary. Move the block to a def instead and give it a compelling name.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/hover/ScalaHoverImpl.scala
```diff
@@ -116,7 +116,21 @@ class ScalaHoverImpl extends ITextHover with ITextHoverExtension with ITextHover
...
+            val gsym = sym.getterIn(sym.owner)
+            val ssym = sym.setterIn(sym.owner)
+            if( gsym != NoSymbol && ssym != NoSymbol)
+              Option("var " + sym.localName + ":" +gsym.tpe.resultType)
```
sschaef: `localName` can return values with a space at the end (I forgot why). Please trim it.

sschaef: Replace `gsym.tpe.resultType` with `compiler.declPrinter(gsym.tpe.resultType)`.

sschaef:
Actually, I think the bug should be fixed in `DeclarationPrinter.defString`. There is also a test: `DeclarationPrinterTest`. This would make the behavior change applicable everywhere and not just in the hover. Once it is fixed there, no further changes need to be done in the hover.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/hover/ScalaHoverImpl.scala
andrey-ilinykh: But I need gsym later. Simply, I don't want to calculate it  twice.

andrey-ilinykh:
I agree, DeclarationPrinter is the best place to fix. Let me look into it.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/hover/ScalaHoverImpl.scala
andrey-ilinykh: What is the difference?

PR #1003, opened 2015-09-10
Step Message Out enhancement

wpopielarski:
changes current 'first met launched' bang operation functionality
into stepping. So clicking on bang button causes:
 - step over in case when 'tell (bang)' method is not found in
   current line
 - run and stop in implemented 'receive' method otherwise

### DIFF
--- a/org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/async/AsyncUtils.scala
+++ b/org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/async/AsyncUtils.scala
@@ -1,6 +1,6 @@
 package org.scalaide.debug.internal.async
 
-import scala.collection.JavaConverters._
+import scala.collection.JavaConverters.asScalaBufferConverter
 
 import org.scalaide.debug.internal.model.JdiRequestFactory
 import org.scalaide.debug.internal.model.ScalaDebugTarget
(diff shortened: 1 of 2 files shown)


sschaef:
Feature changes should have a ticket, otherwise when a release is done I have to go through the commits manually and find out what changed. Please create a ticket for this change too and mention it in the commit message.

sschaef:
Beside from the fact that I would like to have a ticket, LGTM.

sschaef:
The hover text of the new icon says "step until a message is received". This is no longer the case, would you please update the message?

sschaef:
I can't think of a really good description of the hover right now. Maybe "step over and skip execution context"?

wpopielarski:
ok, I got it

2015-09-11 13:09 GMT+02:00 Simon Schäfer <notifications@github.com>:

wpopielarski:
hmm... I ponder a little on tooltip text and looks like Iulian's is best I found. It's concise and saying exactly what this action is doing. I know that 'step over' is a kind of combined trait or behavior (or event a side effect) but it's done for the sake of usability and constraint of debugging process (it's hard to go back into the past).

sschaef:
We can still find a better tooltip description for the final release. Let's not block the RC3 because of this.

wpopielarski:
squashed and ticket created and added to commit

sschaef:
LGTM

PR #1004, opened 2015-09-11
fix var tooltip message (#1001007)

andrey-ilinykh:


wpopielarski:
I just glimpsed at code and I think Simon will confirm, that formatting could be made better :). Actually it is really better place to control of tooltip content.

sschaef:
add to whitelist

org.scala-ide.sdt.core.tests/src/org/scalaide/core/pc/DeclarationPrinterTest.scala
```diff
@@ -142,5 +142,68 @@ class DeclarationPrinterTest {
...
+    runTest("""
+          |var a/**/:Int=4
+          |def f():Unit = {
+          |  println(a)
+          |}
+          |""".stripMargin,
+          "var a: Int")
+  }
+
+  private def runTest(in: String, expected: String): Unit = {
```
sschaef: Could you please rename this method to something like `runTestWithMarker`? This makes it easier to understand the difference between the two `runTest` methods.

sschaef:
As already said, the formatting is a bit off. Especially the whitespace needs to be improved. Would you please run the "format selection" action on a selection of your contributed code?

sschaef:
The commit message needs to be improved too. See [this section](https://github.com/scala-ide/scala-ide/blob/master/CONTRIBUTING.md#creating-commits-and-writing-commit-messages). The ticket needs to be mentioned as `Fix #1001007` in the body and not in the title.

sschaef:
Superseded by #1005

PR #1005, opened 2015-09-11
Fix tooltip message for 'var's

andrey-ilinykh:
Distinguish var from def by checking if the owner has both a setter and
a getter with the same name

Fix #1001007

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/core/compiler/DeclarationPrinter.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/core/compiler/DeclarationPrinter.scala
@@ -78,13 +78,21 @@ abstract class DeclarationPrinter extends HasLogger {
       else nonAccess + " " + access
     }
 
-    val name = if (sym.isConstructor) sym.owner.decodedName else sym.nameString
-    compose(
-      // TODO: Annotations
-      flagString(sym, flagMask & sym.flags),
-      if (showKind) sym.keyString else "",
-      sym.varianceString + name + infoString(seenAs) // don't force the symbol
-      )
+    val gsym= sym.getterIn(sym.owner)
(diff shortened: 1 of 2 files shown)


sschaef:
The formatting is still not perfect (if it had changed at all), but I don't want to be too pedantic. Therefore, LGTM.

sschaef:
Thanks

org.scala-ide.sdt.core/src/org/scalaide/core/compiler/DeclarationPrinter.scala
```diff
@@ -78,13 +78,21 @@ abstract class DeclarationPrinter extends HasLogger {
...
-      // TODO: Annotations
-      flagString(sym, flagMask & sym.flags),
-      if (showKind) sym.keyString else "",
-      sym.varianceString + name + infoString(seenAs) // don't force the symbol
-      )
+    val gsym= sym.getterIn(sym.owner)
+    def hasGetterSetter(sym:Symbol)= gsym != NoSymbol && sym.setterIn(sym.owner) != NoSymbol
+
+    if(hasGetterSetter(sym)){
+      compose("", "var",sym.localName.toString.trim+": "+gsym.tpe.resultType)
```
dragos: I don't think this is working very well, since modifiers are eluded and hardcoded to `""`. What about `private/protected` vars?. Also, the type should be printed using the same calls as in the original code. I think the proper way would be to use the getter symbol for all of these. For an example about the type:

```
trait Foo[T] {
  var x: T = _
}

class C extends Foo[Int] {
  prinntl(x) // x: Int or x: T?
}

andrey-ilinykh:
Yes, I agree. It is not perfect. I hardcoded modifiers to "" because they are wrong any way. In simple case
class Foo{
  val x:Int = 100
}
tooltip is "private[this] val x: Int"

dragos:
@andrey-ilinykh I think there is at least one symbol (not the field, but probably the getter field) that has the correct flags. We should use those.

PR #1006, opened 2015-09-14
Prevent NPE in SbtInputs

sschaef:
`getLocation` can return null in case the underlying container is
invalid. However, I have no idea in which case it can become invalid,
therefore we just throw another exception instead of the NPE in order to
make it easier for users to find the problematic container.

Fixes #1002544


dragos:
If we don't have a reproducible, it's hard to know what to do. Would it make more sense to stop the build (since we have no idea where sources are) and create a problem marker, so at least the user is warned?

sschaef:
Instead of passing null, an exception with a better error message is now thrown.

wpopielarski:
LGTM

dragos:
LGTM

PR #1008, opened 2015-09-20
Make Scala IDE build with Eclipse Mars.

dragos:


dragos:
Obviously, this won't build on Luna anymore... so we can merge it once we take a decision to move development to Mars.

sschaef:
I worked on something similar and I even updated the maven build to handle both Luna and Mars builds: https://github.com/scala-ide/scala-ide/compare/master...sschaef:t1002530 The only last missing piece (as I saw it) was that I didn't find out which weaving version to use for Mars.

dragos:
I remember now. Not sure exactly how to find out that version, but there must be a way :)

Do you know if Mars binaries run on Luna? If yes, I'd keep the build simple and not add the profile and different source folders.

dragos:
Let's close this and get yours in.

mkurz:
When downloading Scala IDE from the [website](http://scala-ide.org/download/sdk.html) it is still based on Eclipse 4.4 (Luna).
When will you switch the base to 4.5 (Mars)?

sschaef:
I hadn't the time to make our build compatible with Mars lately. Also, since in my eyes Mars is not an improvement over Luna, I was also reluctant to do the move. I expect to be able to do the transition in January. You can install Scala IDE on top of Mars in the meantime manually by yourself.

rkrzewski:
I confirm that Scala IDE can be installed and works fine with Eclipse Mars and even Neon :)

mkurz:
Thanks, I know I can install it on top of Mars right now.
I was just curious, because when using Luna we suffer from an annoying bug which got fixed in Mars.

PR #1009, opened 2015-09-21
Improvements on Hover display

dragos:
Show correct modifiers for vals/vars.

@andrey-ilinykh since you recently fixed some bugs in this area, it'd be great if you could have a look at these changes.

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/core/compiler/DeclarationPrinter.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/core/compiler/DeclarationPrinter.scala
@@ -78,21 +78,23 @@ abstract class DeclarationPrinter extends HasLogger {
       else nonAccess + " " + access
     }
 
-    val gsym= sym.getterIn(sym.owner)
-    def hasGetterSetter(sym:Symbol)= gsym != NoSymbol && sym.setterIn(sym.owner) != NoSymbol
+    val gsym = sym.getterIn(sym.owner)
+    val hasGetterSetter = (gsym ne NoSymbol) && (sym.setterIn(sym.owner) != NoSymbol)
 
-    if(hasGetterSetter(sym)){
-      compose("", "var",sym.localName.toString.trim+": "+gsym.tpe.resultType)
-    } else{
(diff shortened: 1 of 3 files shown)


sschaef:
I just tested this PR with the following tests:
```scala
package abc

trait X {
  val x1 = 0
  private val x2 = 0
  private[this] val x3 = 0
  private[abc] val x4 = 0 // broken
  protected val x5 = 0
  protected[this] val x6 = 0
  protected[abc] val x7 = 0 // broken

  var y1 = 0
  private var y2 = 0
  private[this] var y3 = 0
  private[abc] var y4 = 0 // broken
  protected var y5 = 0
  protected[this] var y6 = 0
  protected[abc] var y7 = 0 // broken
}
```

`protected[abc]` is shown as `protected[package abc]`. I think that it would be better to not show the `package` addition.
`private[abc]` isn't shown at all.

org.scala-ide.sdt.core/src/org/scalaide/core/compiler/DeclarationPrinter.scala
```diff
@@ -78,21 +78,23 @@ abstract class DeclarationPrinter extends HasLogger {
...
-       val name = if (sym.isConstructor) sym.owner.decodedName else sym.nameString
-       compose(
-         // TODO: Annotations
-         flagString(sym, flagMask & sym.flags),
-         if (showKind) sym.keyString else "",
-          sym.varianceString + name + infoString(seenAs) // don't force the symbol
-        )
-    }
+    val gsym = sym.getterIn(sym.owner)
+    val hasGetterSetter = (gsym ne NoSymbol) && (sym.setterIn(sym.owner) != NoSymbol)
```
sschaef: The first time you use `ne` and the second time it is `!=`. ;)

dragos: :) will fix

dragos:
Thanks for the additional tests, I'll look into those as well!

sschaef:
Man, I completely forgot about about that PR. LGTM

sschaef:
I'm stupid, I merged even though some tests fail. Going to fix it...

PR #1010, opened 2015-09-21
Use the declaration printer in “Show Type".

dragos:


dragos:
This shows `ListBuffer[Outer.Inner]` instead of `scala.collection.mutable.ListBuffer[org.scalaide.ui.internal.actions.TypeOfExpressionProvider.Outer.Inner]`

sschaef:
LGTM

PR #1011, opened 2015-09-23
Consider "surround block" auto edit only for default partition

sschaef:
We want that this auto edit only works for the default partition because
the other partitions don't contain normal Scala code and therefore this
auto edit isn't very useful there.

Fixes #1002553

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/extensions/autoedits/SurroundBlock.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/extensions/autoedits/SurroundBlock.scala
@@ -1,6 +1,7 @@
 package org.scalaide.extensions
 package autoedits
 
+import org.eclipse.jface.text.IDocument
 import org.eclipse.jface.text.IRegion
 import org.scalaide.core.text.Add
 import org.scalaide.core.text.Replace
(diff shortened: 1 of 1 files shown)


dragos:
LGTM! Thanks, @sschaef

wpopielarski:
I saw code but I'm wonder why this change concerns SurroundBlock only.

2015-09-24 12:37 GMT+02:00 Iulian Dragos <notifications@github.com>:

sschaef:
On 09/24/2015 01:23 PM, Wiesław Popielarski wrote:
> I saw code but I'm wonder why this change concerns SurroundBlock only.
I don't understand, can you elaborate a little bit more?

wpopielarski:
no problem, so there is a bunch of objects extending AutoEditSetting. The
default of partitions is just empty set and only SurroundBlockSettings
alters it. So my question is why SurroundBlock needs it and other not.

2015-09-24 14:11 GMT+02:00 Simon Schäfer <notifications@github.com>:

sschaef:
On 09/24/2015 02:45 PM, Wiesław Popielarski wrote:
> no problem, so there is a bunch of objects extending AutoEditSetting. The
> default of partitions is just empty set and only SurroundBlockSettings
> alters it. So my question is why SurroundBlock needs it and other not.
If no partitions are chosen, the auto edit is enabled for all existing 
partitions. In the case at hand we only want to enable it for the 
default partition. To which partition an auto edit should be applied 
depends on the concrete auto edit, therefore this change only affects 
the auto edit at hand.

PR #1012, opened 2015-09-24
Fix non working fading packages feature in Async Debug View

sschaef:
The delimiter sign has changed. We use the constant instead of the
broken hardcoded one.

Fixes #1002549


wpopielarski:
LGTM

PR #1014, opened 2015-10-09
New Outline View

andrey-ilinykh:
This is initial version of OutlineView based on parsed tree. I created one more reconciling strategy (ScalaOutlineReconcilingStrategy) which updates model for the TreeView. Model is built by ModelBuilder. This job is done on reconciler thread, then result is propagated to UI thread.


org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ModelBuilder.scala
```diff
@@ -0,0 +1,257 @@
...
+          //logger.info("ModuleDef "+name)
+          val ch =  ObjectNode(name.toString(), parent)
+          setPos(ch, t)
+          ch.setFlags(mods.flags)
+          parent.addChild(ch)
+          t.children.foreach(x => updateTree(ch, x))
+
+        }
+        case DocDef(comment: DocComment, definition: Tree) => updateTree(parent, definition)
+        case _ => //logger.info("_"+t.getClass)
```
dragos: I think you should also treat `TypeDef` (for type aliases)

dragos: Actually, the current outline view also shows import statements, so it might be good to show them too (at least the top-level ones)

dragos:
Overall it looks good, thanks a lot for working on this. It's really great to have this part of front-facing editor being revamped on solid grounds!

I would like to give it a try and play with it. I just had a glance at how the trees are used. It looks good. There might be some Scala tweaks that could be improved, but the approach looks fine. @sschaef or @wpopielarski, it'd be good if you guys could have a look as well.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ModelBuilder.scala
```diff
@@ -0,0 +1,257 @@
...
+        //        case AppliedTypeTree(tpt, args) =>logger.info("AppliedTypeTree "+ tpt+", "+args)
+        //        case Apply(fun, args) => logger.info("Apply "+ fun)
+        case DefDef(mods, name, tparamss, vparamss, tpt, rsh) =>
+          if ("<init>" != name.toString) {
+            def typeList = {
+              vparamss.map { x => x.map { v => v.tpt.toString() } }
+            }
+            val ch =  MethodNode(name.toString(), parent, typeList)
+            //logger.info("DefDef " + name + ", modes=" + mods + ", tpt=" + tpt.hasSymbolField + ", " + rsh)
+            ch.returnType = if (tpt.hasSymbolField) Some(tpt.toString) else None
```
dragos: I wonder why this test was this needed

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ScalaOutlineReconcilingStrategy.scala
```diff
@@ -0,0 +1,49 @@
...
+        val oldRoot = sop.getInput
+        icUnit.scalaProject.presentationCompiler.apply(comp => {
+          val rootNode =ModelBuilder.buildTree(comp, icUnit.sourceMap(icuEditor.getViewer.getDocument.get.toCharArray()).sourceFile)
+          val delta = if(oldRoot!= null) oldRoot.diff(rootNode) else null
+          icuEditor.getEditorSite.getShell.getDisplay.asyncExec(new Runnable() {
+            def run = {
+              if(delta eq null)
+                sop.setInput(rootNode)
+              else
+                sop.update(delta)
```
dragos: Personally, given that the parser is really fast, and we need to parse again anyway from scratch anyway, I wouldn't go through the trouble of computing a `delta`. Unless there's some other benefit, like UI flickering if the whole tree is replaced (I could see that as being very annoying). Is that the reason?

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ModelBuilder.scala
andrey-ilinykh: Sure. It is easy

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ModelBuilder.scala
andrey-ilinykh: It is just a synthetic symbol (constructor) which doesn't hold any new information then a class node itself.  And name (<init>) looks weird. But I just realized that the current implementation has a filter which removes synthetic symbols. Probably I have to do the same.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ScalaOutlineReconcilingStrategy.scala
andrey-ilinykh: Yes, the tree could be quite big - thousands of nodes. Each node translates to OS widget. When you replace the tree you have to dispose all of them first then and initialize again. It could be expensive and should be done on UI thread. Another problem- the viewer has a state. User can expand/collapse some nodes. When you replace data you loose this state. To restore it you have to track all expand/collapse actions. It seems to be too complicated. I prefer to calculate nodes which get changed.

org.scala-ide.sdt.core/src/org/scalaide/core/internal/ScalaPlugin.scala
```diff
@@ -123,6 +123,8 @@ class ScalaPlugin extends IScalaPlugin with PluginLogConfigurator with IResource
   /** A LRU cache of class loaders for Scala builders */
   lazy val classLoaderStore: FixedSizeCache[IScalaInstallation,ClassLoader] = new FixedSizeCache(initSize = 2, maxSize = 3)
 
+  /** eventually scala plugin should have its own image description registry */
```
sschaef: This should be a TODO, not the Scaladoc of this value.

org.scala-ide.sdt.core/src/org/scalaide/ui/ScalaImages.scala
```diff
@@ -37,4 +37,8 @@ object ScalaImages {
   val DESC_SETTINGS_OBJ: ImageDescriptor = OSGiUtils.getImageDescriptorFromCoreBundle("/icons/full/obj16/settings.gif")
 
   val ADD_METHOD_PROPOSAL: Image = JavaPluginImages.DESC_MISC_PUBLIC.createImage()
+  val PRIVATE_VAR :ImageDescriptor = JavaPluginImages.DESC_FIELD_PRIVATE
```
sschaef: Formatting is broken (colon, no space before, space after). Please do format action on your newly introduced code.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ModelBuilder.scala
```diff
@@ -0,0 +1,257 @@
...
+abstract class NodeKey
+
+abstract class Node(val isLeaf: Boolean) extends HasModifiers {
+  def name: String
+  def parent: ContainerNode
+  var start: Int = 0
+  var end: Int = 0
+  def displayName: String
+  def key: NodeKey
+  def <= (that:Node):Boolean={
```
sschaef: In Scala equality methods should not be overwritten. Use an `Ordering` instead.

sschaef: Also, you shouldn't write such code. A comparison is supposed to be a side effect free operation. But yours is not.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ModelBuilder.scala
```diff
@@ -0,0 +1,257 @@
...
+    start= that.start
+    end = that.end
+    val f = flags
+    flags = that.flags
+    f != that.flags
+  }
+}
+
+abstract class ContainerNode(parent: ContainerNode) extends Node(false) {
```
sschaef: This parent should override super.parent

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ModelBuilder.scala
```diff
@@ -0,0 +1,257 @@
...
+           case None =>
+             hasNew =true
+             children += (p._1-> p._2)
+        }
+        )
+        if(children.size != cn.children.size)
+          hasNew = true
+        if(hasNew )
+          toUpdate += cn
+        cn.children = children
```
sschaef: A diff method shouldn't have a side effect either.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ModelBuilder.scala
```diff
@@ -0,0 +1,257 @@
...
+  import scala.reflect.internal.Flags._
+  private[editor] var flags = 0L
+  def setFlags(f: Long) = {
+    flags = f
+  }
+  def isPrivate = (flags & PRIVATE) != 0
+  def isProtected = (flags & PROTECTED) != 0
+  def isTrait = (flags & TRAIT) != 0
+}
+abstract class NodeKey
```
sschaef: NodeKey, Node and ContainerNode should be sealed.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ModelBuilder.scala
```diff
@@ -0,0 +1,257 @@
...
+      case _ => b
+    }
+  }
+}
+
+
+
+object ModelBuilder extends HasLogger {
+
+  def buildTree(comp: IScalaPresentationCompiler,src:SourceFile):RootNode={
```
sschaef: I can't say that I like the mutable nature of the tree. I would expect a much simpler implementation of this method with an immutable tree.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ModelBuilder.scala
sschaef: Indeed, just checking for `"<init>"` is not enough. There are other synthetic members (mainly the stuff that contains $ signs).

sschaef:
`def +++ =  0` is shown as `$plus$plus$plus`. There is a class in the compiler which can unmangle internal names but I forgot how it is called.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ScalaContentProvider.scala
```diff
@@ -0,0 +1,86 @@
+package org.scalaide.ui.internal.editor
+import org.eclipse.jface.viewers.ITreeContentProvider
+import org.eclipse.jface.viewers.Viewer
+
+
+
+
+/**
+ * @author ailinykh
+ */
```
sschaef: Please don't put `@author` tags anywhere. There is an `AUTHORS` file in the root dir.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ScalaContentProvider.scala
```diff
@@ -0,0 +1,86 @@
...
+class ScalaContentProvider extends ITreeContentProvider {
```
sschaef: `ScalaContentProvider` is not a good name for this class, since it is only related to the Outline View.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ScalaContentProvider.scala
```diff
@@ -0,0 +1,86 @@
...
+  /**
+   * Called when the input changes
+   *
+   * @param arg0
+   *            the viewer
+   * @param arg1
+   *            the old input
+   * @param arg2
+   *            the new input
+   */
```
sschaef: Scaladoc on overridden members can be removed.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ScalaLabelProvider.scala
```diff
@@ -0,0 +1,114 @@
...
+
+
+ */
+import org.eclipse.jface.viewers.ILabelProvider
+import java.io.FileInputStream
+import org.eclipse.swt.graphics.Image
+import org.eclipse.jface.viewers.ILabelProviderListener;
+import org.scalaide.core.internal.ScalaPlugin
+
+class ScalaLabelProvider extends ILabelProvider{
```
sschaef: Again, bad name.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ScalaLabelProvider.scala
```diff
@@ -0,0 +1,114 @@
...
+   *
+   * @param arg0
+   *            the node
+   * @return String
+   */
+  override def  getText( o:Object):String ={
+    def renderArgList(sb:StringBuilder, args:List[List[String]])={
+      args.foreach(list =>{
+        sb.append("(")
+        list.foreach { s => sb.append(s); sb.append(",") }
```
sschaef: I would like to see a space after the comma.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ScalaLabelProvider.scala
```diff
@@ -0,0 +1,114 @@
...
+    o match {
+      case c:PackageNode => c.displayName+ " "+u
+      case c:ClassNode => c.displayName+ " "+u
+      case c:ValNode => c.displayName + " "+u
+      case c:VarNode => c.displayName + " "+u
+      case c:ObjectNode => c.displayName+ " "+u
+      case c:MethodNode => {
+        val sb = new StringBuilder
+        if(!c.argTypes.isEmpty)
+          renderArgList(sb, c.argTypes)
```
sschaef: param names should be shown.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ScalaLabelProvider.scala
```diff
@@ -0,0 +1,114 @@
...
+        c.displayName+sb.toString() + c.returnType.map(":"+_).getOrElse("")+" "+u
```
sschaef: a space should be shown after the colon

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ScalaLabelProvider.scala
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  /**
+   * Gets the text to display for a node in the tree
+   *
+   * @param arg0
+   *            the node
+   * @return String
+   */
+  override def  getText( o:Object):String ={
```
sschaef: I think most of the implementation of this method is needless. We already have `DeclarationPrinter`. It should be used here too.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ScalaOutlinePage.scala
```diff
@@ -0,0 +1,134 @@
...
+import org.eclipse.ui.handlers.CollapseAllHandler
+import org.eclipse.jface.action.Action
+import org.eclipse.jdt.internal.ui.JavaPluginImages
+import org.eclipse.ui.PlatformUI
+import org.eclipse.jdt.internal.ui.IJavaHelpContextIds
+import org.eclipse.jdt.internal.ui.JavaPlugin
+import org.eclipse.jdt.internal.ui.javaeditor.JavaEditorMessages
+import org.scalaide.logging.HasLogger
+
+class ScalaOutlinePage(val fEditor: ScalaSourceFileEditor) extends ContentOutlinePage with HasLogger {
```
sschaef: The implementation of this class looks like copied from JDT. If so, please add a remark to the Scaladoc that mentions the origin of this code. Also, spend some further time in cleanup. Semi colons should be removed, code should be formatted, members need to be sorted, an empty line between members would be nice.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ScalaOutlineReconcilingStrategy.scala
```diff
@@ -0,0 +1,49 @@
...
+            }
+          })
+        })
+      }
+    } catch {
```
sschaef: Which exception can be thrown here?

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ScalaOutlineReconcilingStrategy.scala
```diff
@@ -0,0 +1,49 @@
...
+
+  override def reconcile(partition: IRegion): Unit = {
+    val sop = icuEditor.getOutlinePage
+    try {
+      if (sop != null) {
+        val oldRoot = sop.getInput
+        icUnit.scalaProject.presentationCompiler.apply(comp => {
+          val rootNode =ModelBuilder.buildTree(comp, icUnit.sourceMap(icuEditor.getViewer.getDocument.get.toCharArray()).sourceFile)
+          val delta = if(oldRoot!= null) oldRoot.diff(rootNode) else null
+          icuEditor.getEditorSite.getShell.getDisplay.asyncExec(new Runnable() {
```
sschaef: There exists `DisplayThread.asyncExec`.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ScalaOutlineReconcilingStrategy.scala
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  override def reconcile(partition: IRegion): Unit = {
+    val sop = icuEditor.getOutlinePage
+    try {
+      if (sop != null) {
+        val oldRoot = sop.getInput
+        icUnit.scalaProject.presentationCompiler.apply(comp => {
+          val rootNode =ModelBuilder.buildTree(comp, icUnit.sourceMap(icuEditor.getViewer.getDocument.get.toCharArray()).sourceFile)
+          val delta = if(oldRoot!= null) oldRoot.diff(rootNode) else null
```
sschaef: In Scala we prefer to wrap potential nulls in Option: `val sop = Option(icuEditor.getOutlinePage)`

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ScalaSourceFileEditor.scala
```diff
@@ -95,9 +98,7 @@ class ScalaSourceFileEditor
     }
 
     override def aboutToBeReconciled() = ()
-    override def reconciled(ast: CompilationUnit, forced: Boolean, progressMonitor: IProgressMonitor) = {
-      getInteractiveCompilationUnit() match {
-        case scu: ScalaCompilationUnit => exts foreach { ext =>
+    override def reconciled(ast: CompilationUnit, forced: Boolean, progressMonitor: IProgressMonitor) = {      getInteractiveCompilationUnit() match {        case scu: ScalaCompilationUnit => exts foreach { ext =>
```
sschaef: Formatting is broken on this line

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ScalaSourceFileEditor.scala
```diff
@@ -439,7 +440,35 @@ class ScalaSourceFileEditor
       else
         super.requestWidgetToken(requester, priority)
   }
+  private var outlinePage:ScalaOutlinePage = null
```
sschaef: The outline page also needs to work for `ScalaClassFileEditor`. Please move this to `ScalaCompilationUnitEditor` to handle both.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ScalaSourceFileEditor.scala
```diff
@@ -439,7 +440,35 @@ class ScalaSourceFileEditor
...
+  def getOutlinePage = outlinePage
+  override def getAdapter( required:Class[_]):AnyRef={
```
sschaef: When is the outline view not available? In case it is closed? Is `getAdapter` never called in this case?

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ScalaSourceFileEditor.scala
```diff
@@ -439,7 +440,35 @@ class ScalaSourceFileEditor
...
+  private def createScalaOutlinePage:ScalaOutlinePage ={
+    val sop = new ScalaOutlinePage(this)
+    sop
+  }
+  override def doSelectionChanged(selection:ISelection)={
+    selection match {
+      case ss:IStructuredSelection =>
+        if(!ss.isEmpty()){
+          ss.getFirstElement match {
+            case n:Node => setHighlightRange(n.start, n.end - n.start, true);
```
sschaef: Scaladoc for what this is doing would be nice.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ScalaSourceFileEditor.scala
sschaef: Overall, please try to move all of this outline stuff to a trait called `OutlineEditorExtension`, similar to how it is done in `MarkOccurrencesEditorExtension` and mix it into the `ScalaCompilationUnitEditor`.

sschaef:
Done with review. My main concerns:

- Consider better formatting of your source code -> auto format
- `ModelBuilder` should be immutable
- All new files should be moved to the package `org.scalaide.ui.internal.editor.outline`.
- Do not do any type printing by yourself. Use `DeclarationPrinter`.

I found some bugs in the printing logic, I guess the moment when you swith to `DeclarationPrint`, these problems are gone:
```
object Test {
  def +++ = 0 // wrong, should be shows as +++
  def f1 = 0
  def f2: Int = 0
  def f3(i: Int) = 0
  def f4(i: Int): Int = 0
  def f5(i: Int, j: Int) = 0
  def f6(i: Int, j: Int): Int = 0
  def f7(i: Int, j: Int)() = 0
  def f8(i: Int, j: Int)(k: Int): Int = 0
  def f9(i: Int, j: Int)(implicit k: Int): Int = 0 // wrong, no implicit shown
  def f10(i: Int)(j: Int)(implicit k: Int) = 0
  def f11(f: Int ⇒ Int) = 0 // wrong, should be shown as Int ⇒ Int
  def f12(t: (Int, Int)) = 0 // wrong, should be shown as (Int, Int)

  trait Outer {
    // wrong, $init$() is shown
    trait Inner
  }
  def f13(o: Outer)(i: o.Inner) = ???

  type Type = Int // wrong, not shown
  def f14(t: Type) = ???
}
```

Most of my points are regarding cleanup. Beside from that, the code works really fine. Good job!

dragos:
Just a couple more points to what @sschaef added:

- type parameters should be also printed
- inner (local) methods should also appear in the Outline (a local function inside a method)
- type members too
- see if the same mechanism can be used for Quick Outline (Ctrl-O).

I'm really looking forward to seeing this in master, it's a major improvement!

andrey-ilinykh:
Ok. I just pushed my latest changes.

andrey-ilinykh:
one more push, should fix build

org.scala-ide.sdt.core.tests/src/org/scalaide/core/ui/OutlineModelTest.scala
```diff
@@ -0,0 +1,95 @@
...
+
+  @Test
+  def testFunArg(): Unit = {
+    runTest("""package pack
+               class Foo{
+                def p2= {""}
+                def p:Int => Int ={i => i*i}
+                def a(f:Int =>MyType)(implicit a:Int,  b:Long) = {}
+                def a1(f:(Int, Long) =>String) = {}
+                def a2(f:((Int, Double)) =>String) = {}
```
sschaef: this should also test call-by-name arguments

sschaef: Another test idea: return a tuple in the function.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/outline/ModelBuilder.scala
```diff
@@ -0,0 +1,417 @@
...
+}
+sealed abstract class NodeKey
+
+sealed abstract class Node(val isLeaf: Boolean, val parent: ContainerNode) extends HasModifiers {
+  def name: String
+  var start: Int = 0
+  var end: Int = 0
+  def displayName = name
+  def key: NodeKey
+  def cp(that: Node): Boolean = {
```
sschaef: What does `cp` mean?

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/outline/ModelBuilder.scala
```diff
@@ -0,0 +1,417 @@
...
+        })
+      if (children.size != cn.children.size)
+        needUpdate = true
+      if (needUpdate)
+        toUpdate += cn
+      cn.children = children
+    }
+
+    def visitNode(n: Node, n1: Node): Boolean = {
+      val b = n.cp(n1)
```
sschaef: This has side effects. Can you please make them clean by documenting it in Scaladoc. It may also be useful to rename the `diff` method to make the side effects clear.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/outline/ScalaOutlineReconcilingStrategy.scala
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  override def reconcile(partition: IRegion): Unit = {
+    val sop = Option(icuEditor.getOutlinePage)
+    if (!sop.isEmpty) {
+      val oldRoot = sop.get.getInput
+      icUnit.scalaProject.presentationCompiler.apply(comp => {
+        val rootNode = ModelBuilder.buildTree(comp, icUnit.sourceMap(icuEditor.getViewer.getDocument.get.toCharArray()).sourceFile)
+        val delta = if (oldRoot != null) oldRoot.diff(rootNode) else null
+        DisplayThread.asyncExec(
```
sschaef: I think it is better to place the `asyncExec` outside of the PC thread.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/outline/ModelBuilder.scala
```diff
@@ -0,0 +1,417 @@
...
+              setPos(in, t)
+            }
+          }
+
+        case _ => //logger.info("_"+t.getClass+", p="+ t.toString)
+      }
+    }
+    val rootNode = new RootNode
+    val t = comp.parseTree(src)
+    comp.asyncExec(updateTree(rootNode, t)).get
```
dragos: Is this call needed? Do you call any methods on `Symbol` or `Type` inside? If not, there shouldn't be a need to execute this on the compiler thread. Parse trees are the only things that are thread-safe in the compiler. :)

andrey-ilinykh: I just was playing with declPrinter, forgot to remove

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/outline/ScalaOutlineReconcilingStrategy.scala
andrey-ilinykh: this is not PC thread, this is a reconciler thread

andrey-ilinykh:
I added by name parsing.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/outline/ModelBuilder.scala
dragos: :+1:

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/outline/OutlinePageEditorExtension.scala
```diff
@@ -0,0 +1,43 @@
...
+            case _ => super.doSelectionChanged(selection)
+          }
+        }
+    }
+  }
+
+}
+
+object OutlinePageEditorExtension {
+  val iContentOutlinePage = classOf[org.eclipse.ui.views.contentoutline.IContentOutlinePage]
```
sschaef: vals should start with upper case letter. If you import the `IContentOutlinePage`, you could also completely inline this val.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/outline/ModelBuilder.scala
```diff
@@ -0,0 +1,435 @@
...
+  def isSynthetic = (flags & SYNTHETIC) != 0
+  def isImplicit = (flags & IMPLICIT) != 0
+}
+sealed abstract class NodeKey
+
+sealed abstract class Node(val isLeaf: Boolean, val parent: ContainerNode) extends HasModifiers {
+  def name: String
+  var start: Int = 0
+  var end: Int = 0
+  def displayName = name
```
sschaef: I do not really like the `displayName` method. Some of the string conversion is done in here, the rest is done in `ModelBuild.buildTree`. I think it is more maintainable when everything is in one place. What about replacing `displayName` by a big pattern match?

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/outline/ModelBuilder.scala
```diff
@@ -0,0 +1,435 @@
+package org.scalaide.ui.internal.editor.outline
+/**
```
sschaef: This comment should be above the `ModelBuilder` type, otherwise Scaladoc can't find it.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/outline/ModelBuilder.scala
```diff
@@ -0,0 +1,435 @@
...
+  def isPrivate = (flags & PRIVATE) != 0
+  def isProtected = (flags & PROTECTED) != 0
+  def isTrait = (flags & TRAIT) != 0
+  def isParam = (flags & PARAM) != 0
+  def isSynthetic = (flags & SYNTHETIC) != 0
+  def isImplicit = (flags & IMPLICIT) != 0
+}
+sealed abstract class NodeKey
+
+sealed abstract class Node(val isLeaf: Boolean, val parent: ContainerNode) extends HasModifiers {
```
sschaef: Would you please all `Node` types and the traits above to a file called `Nodes`?

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/outline/ModelBuilder.scala
andrey-ilinykh: Actually, Node text rendering is responsibility of a label provider. Right now we use implementation of ILabelProvider, which doesn't care about font and colors. So it was possible to delegate that rendering to the node. But theoretically speaking we may implement more sophisticated rendering using different colors/fonts. That's why I moved this logic to ScalaOutlineLabelProvider.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/outline/OutlinePageEditorExtension.scala
andrey-ilinykh: Honestly I did'n get the part about inlining. Could you explain?

sschaef: On 10/29/2015 06:27 PM, Andrey Ilinykh wrote:
>
> Honestly I did'n get the part about inlining. Could you explain?
>
I meant that you could inline the value into the places where it is 
used. But doesn't really matter.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/outline/ModelBuilder.scala
sschaef: Ok

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/outline/ScalaOutlineReconcilingStrategy.scala
sschaef: You are right. Everything fine here.

sschaef:
I found more things to improve:

- There still exist some symbols, that should be sorted out: `<empty>` (package), `<init>()` (all synthetic ctors), `$init$`, `$init$()`
- multi imports should include a space after a comma. Right now they are shown as `a.b.c.{A,B}`
- `package` declarations at the top should be shown before import declarations. Right now they are shown below them
- local import declarations contain a header `import declarations`, which does have the right location set. If one clicks on the item in the outline view, the cursor jumps to line one. I think it is best to jump to the location of the first entry
- Not sure if that can be solved, but some entries (especially methods), contain an arrow on their left to expand the view, but clicking on it doesn't reveal anything. The arrow shouldn't be shown when no expansion exists.
- `val (t1, t2) = (1, 2)` does show an entry `x$2` beside from `t1` and `t2`. pattern matching seems to create these values but they should be sorted out
- I think it is best to truncate fully qualified names. `akka.actor.ActorSystem` should only be shown as `ActorSystem`.
- You also need to implement the feature for the `ScalaClassFileEditor` otherwise an exception is thrown when one tries to open a class file (because you do a cast in the `ScalaSourceViewerConfiguration.getReconciler` method)
- type parameter lists in methods, types, classes and traits are not shown. `def g[A](a: A) = ???` is shown as `g(a: A)`. I don't think we have to show everything here (i.e no upper, lower or view bounds) but at least the name of the arguments.
- the quick outline doesn't show this new outline view. But don't ask me how one can change that, I don't know where it is configured.

Beside from that, the new outline view looks really good. It is also very fast (it has a small startup time for multi thousand lines of code files but I can live with that). It is not far away from getting my green light.

andrey-ilinykh:
I fixed everything except QuickOutlineView. It will take more time, and we will lose some functionality. Right now QuickOutlineView allows to see inherited members. It is not possible with parsed tree.

sschaef:
On 10/31/2015 03:34 AM, Andrey Ilinykh wrote:
>
> I fixed everything except QuickOutlineView. It will take more time, 
> and we will lose some functionality. Right now QuickOutlineView allows 
> to see inherited members. It is not possible with parsed tree.
>
But isn't that the case for the Outline View too? I think it is ok to 
loose that functionality, the outline view should represent the current 
file on a high level. Beside from that anything semantic in the outline 
view doesn't work reliable anyway so far, therefore I don't see it as a 
loss more. Any semantic features that are reliable implemented can be 
added in future.

sschaef:
I noticed another weird thing about package declarations. If one selects them in the outline view, the editor always shows the selected line at the top, even if the package is already in the view area of the editor. For all other things, the selected line is shown at about a third of the screen when they are not already shown in the view area.

sschaef:
`def x[A, B]: A <:< B = ???` is shown as `x[A, B]: $less$colon$less[A,B]`. You need to put the return types in `reflect.NameTransformer.decode`.

sschaef:
Aha, I found another broke case:
```
new Object {} // shown as $anon
val s1 = new Object {} // ok
def s2 = new Object {} // again shown as $anon inside of s2
```
The anonymous class is shown as $anon. I wouldn't show it at all, since it is no definition when it is not bound to a value. Also, I found an update problem in the outline view. `s1` doesn't have an arrow, because it doesn't contain any members. When I now convert it from a `val` to a `def`, no arrow is shown but it has now a member. I first have to add a param list or change something else to get the arrow back.

org.scala-ide.sdt.core.tests/src/org/scalaide/core/ui/OutlineModelTest.scala
```diff
@@ -0,0 +1,166 @@
...
+import org.scalaide.ui.internal.editor.outline._
+import org.hamcrest.core.IsInstanceOf
+
+object OutlineModelTest extends TestProjectSetup("outline-model") {
+  val unit = scalaCompilationUnit("/pack/Target.scala")
+
+  val compUtils = new CompilerTestUtils(unit)
+}
+
+class OutlineModelTest {
```
sschaef: I just noticed that this test is not run. You need to add it to `TestsSuite`.

org.scala-ide.sdt.core.tests/src/org/scalaide/core/ui/OutlineModelTest.scala
```diff
@@ -0,0 +1,166 @@
+package org.scalaide.core.ui
+
+import org.scalaide.core.testsetup.TestProjectSetup
+import org.scalaide.core.CompilerTestUtils
+import org.junit.Test
+import org.junit.Assert
+import org.scalaide.ui.internal.editor.outline._
+import org.hamcrest.core.IsInstanceOf
+
+object OutlineModelTest extends TestProjectSetup("outline-model") {
```
sschaef: For this it is probably not an improvement but I just want to point it out: You do not need to use `TestProjectSetup` and therefore there is also no need for the `outline-model` project files. Instead, your object can mixin `CompilerSupportTests` and implement the `runTest` method as this:

```scala
private def runTest(str: String, f: RootNode => Unit): Unit = {
  import OutlineModelTest._
  withCompiler { comp =>
    val unit = mkScalaCompilationUnit(str)
    val rn = ModelBuilder.buildTree(comp, unit.sourceMap(str.toCharArray).sourceFile)
    f(rn)
  }
}
```

Only downside is that your package names would have to differ. Instead of just `pack` you could name them in the same way how the test method is named. This is necessary because the compilation units are generated and by using different package names we ensure that there are no clashes between type names etc.

But as I said only a small improvement if an improvement at all.

sschaef:
I'm done with another round of review. I hope I checked every feature Scala has (probably I didn't) and it looks really good now.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/outline/ModelBuilder.scala
```diff
@@ -0,0 +1,242 @@
...
+            renderFunc(sb, b :: tail)
+          case a :: Nil =>
+            sb.setLength(sb.length - 2)
+            sb.append(") => ")
+            renderType(sb, a)
+          case Nil =>
+        }
+      }
+      def renderATT(sb: StringBuilder, tt: Tree, tpt: Tree, args: List[Tree], dp: Boolean): Unit = {
+        tpt.toString match {
```
dragos: I think this would be *much* cleaner by matching against the actual type structure, than calling `toString` and using regular expressions to de-structure. All the structured information is already in the tree.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/outline/ModelBuilder.scala
```diff
@@ -0,0 +1,242 @@
...
+              ValNode(name.decodedName.toString(), parent, if (tpt.isEmpty) None else Some(showType(tpt)))
+            else
+              VarNode(name.decodedName.toString(), parent, if (tpt.isEmpty) None else Some(showType(tpt)))
+            setPos(ch, t)
+            ch.setFlags(mods.flags)
+            parent.addChild(ch)
+
+          }
+
+        case DefDef(mods, name, tparamss, vparamss, tpt, rsh) =>
```
dragos: I don't see recursing inside the right-hand side of this method. It would be really useful to show local methods (the current outline view does that).

dragos: Hm, for some reason it seems work :)

dragos:
A couple of things I noticed:

- imports that hide a name, like `import scala.{Predef => _}` are shown as `import scala.Predef`
- parenthesis around functions are lost, so the function type is displayed wrongly. For example, `def bar(x: (Int => Int) => Int)` is displayed as `Int => Int => Int`, which is a different type.
- type bounds are not displayed at all. For instance, `def test[T <: Ordered[T]]` is displayed as `test[T]`. But I actually like this (the outline view is less cluttered), so maybe this could be added later (or as an option).

sschaef:
> type bounds are not displayed at all. For instance, def test[T <: Ordered[T]] is displayed as test[T]. But I actually like this (the outline view is less cluttered), so maybe this could be added later (or as an option).

My idea was to skip type bounds because they can be quite complex (especially with structural types). But sure, if @andrey-ilinykh wants to implement them I have nothing against them.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/outline/ModelBuilder.scala
sschaef: the recursion for `rsh` is in the last line of this block.

dragos:
@sschaef I agree with you. It doesn't have to be in this PR.

One more thing: when selecting a member, the old outline used to highlight the name of the member, not just move to that line. I think that's good behaviour, and probably easy to implement.

andrey-ilinykh:
I just pushed QuickOutlineView implementation. Unfortunately, AbstractInformationControl has private methods written against Java model, so I had to copy paste this class. Some minor details are dropped.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/outline/ModelBuilder.scala
andrey-ilinykh: Yes, I agree. It was shortcut. It seems to work, but then I realized toString method could be dangerous. I see some PC complains about using compiler from wrong thread. Probably toString method try to do something with symbols. I will look into it.

andrey-ilinykh:
Ok. I fixed everything, but there is one concern. Right now we show both nested functions and local variables (I mean variables on the stack). Variables are pretty much useless, should I keep them? If we decide to show them, then probably icon should be different.

sschaef:
Idents in backticks are still not handled correctly. Example:

```scala
package a.b.c // shown only as "c". I think it is better to show it as "a.b.c"

class Test {

  def `=_=`(i: Int) = 0
  def f(i: Int): `A.B` = ??? // shown only as "B"
  def g(i: Int): `A.B`.type = ??? // shown only as "B.type"
  def h(i: Int): `A B` = ??? // shown as "A B"
}

class `A.B`
object `A.B`
class `A B`
```

You still seem to split somewhere on a dot but you should split on the last dot coming from right. Also, idents in backticks don't have the correct selection. If I select `=_=` in the outline view, the backticks need to be selected too. Also have a look to the package declaration. It is not critical, but I would show the entire package as it is written to the sources.

sschaef:
Actually, forget what I wrote about packages. It is ok as it is right now. Otherwise it would just get confusing for multiple package declarations.

sschaef:
On 11/09/2015 06:09 PM, Andrey Ilinykh wrote:
>
> Ok. I fixed everything, but there is one concern. Right now we show 
> both nested functions and local variables (I mean variables on the 
> stack). Variables are pretty much useless, should I keep them? If we 
> decide to show them, then probably icon should be different.
>
I would let it be as it is right now. Variables are also part of the 
structure and I don't think they are particularly noisy. I don't know 
where to get a better icon from. If you can make one by yourself, just 
do it. Otherwise the one that is used right now is also ok.

sschaef:
A potential improvement (but unnecessary for this PR): If one searches in the Quick Outline View, it would be great if the matching idents would highlight the matched part.

sschaef:
There is still a small problem about highlighting: When a trait is selected in the outline view, it is shown at the top of the file, instead of the first third of the screen as it is the case for other things.

andrey-ilinykh:
I believe the package node was rendering incorrectly, I fixed it as well.

sschaef:
Alright, I lost track about everything that was mentioned, the only unsolved issue I can see right now is that traits selections jump to the beginning of the file. But this is just a minor annoyance and can be fixed at a later time. Therefore, LGTM.

@andrey-ilinykh Do you have anything left you want to do for this PR?

andrey-ilinykh:
No, I don't.

sschaef:
Thanks for the contribution Andrey!

sschaef:
I closed nearly all open tickets I could find about the old outline view. Also, I created a few new ones, they are tagged with `outline` and can be seen here: https://scala-ide-portfolio.assembla.com/spaces/ae55a-oWSr36hpeJe5avMc/tickets?tickets_report_id=u1200023&tag_id=1086853

If you are interested in doing further work for the Outline View, Andrey, we would welcome further contributions!

PR #1015, opened 2015-10-10
Do not call refactoring.prepare outside of PC thread

sschaef:
This fixes the race condition described in the ticket.

Fixes #1002562

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/refactoring/internal/rename/Rename.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/refactoring/internal/rename/Rename.scala
@@ -4,6 +4,7 @@ package rename
 import scala.tools.refactoring.analysis.GlobalIndexes
 import scala.tools.refactoring.implementations
 
+import org.scalaide.core.compiler.IScalaPresentationCompiler.Implicits._
 import org.scalaide.refactoring.internal.RefactoringExecutor
 import org.scalaide.refactoring.internal.RefactoringHandler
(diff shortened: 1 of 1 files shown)


sschaef:
retest this please

mlangc:
Looks good and works for me.

org.scala-ide.sdt.core/src/org/scalaide/refactoring/internal/rename/Rename.scala
```diff
@@ -49,7 +50,11 @@ class Rename extends RefactoringHandler {
           new refactoring.FileSelection(source.file, tree, start, end)
         }
 
-        selection map refactoring.prepare flatMap (_.right.toOption) map {
+        val preparation = selection flatMap { selection =>
```
wpopielarski: generally I try to avoid to give same name to option and value but... not sure, semantically it represents same, so ok for me

wpopielarski:
LGTM

org.scala-ide.sdt.core/src/org/scalaide/refactoring/internal/rename/Rename.scala
sschaef: I do exactly the opposite, because I want to hide outer definitions to make it impossible to access the wrong one.

mlangc: +1

PR #1016, opened 2015-10-12
Wrap addImport in asyncExec

sschaef:
It is interesting which API calls access the PC. I wouldn't have
thought that addImport needs access to it too, otherwise this bug may
never have seen the light of birth. Now it is gone and I have to think
once again how to develop an API that prevent such mistakes.

Fixes #1002565


wpopielarski:
LGTM

PR #1017, opened 2015-10-12
Enables Expression Evaluator for remote debugging

wpopielarski:
The project classpath needs to be set for Debugger machine as well.
In other case it is not possible to evaluate instances on Debugger
side and send the output to Debugee machine.

Fixes #1002439


sschaef:
The code looks good. The only problem is that it is not easy to test. I would have to run a patched IDE in the IDE and then spawn an IDE in the test IDE. That is a three level deep spawned IDE.

sschaef:
Ok, I tested it and it works. LGTM

PR #1018, opened 2015-10-13
Do not use system default line delimiter by default

sschaef:
This one is not always correct. Instead, the legal line delimiters have
to be checked and one of them has to be chosen.

Fixes #1002567


wpopielarski:
LGTM

PR #1019, opened 2015-10-13
Explicitly set empty extension name for backup files

sschaef:
On Mac, this parameter is needed, otherwise an error is shown. See this
SO post for more information:

  http://stackoverflow.com/questions/4247068/sed-command-failing-on-mac-but-works-on-linux


sschaef:
This really doesn't need review.

PR #1020, opened 2015-10-15
Add option to preserve wildcard imports during organize imports

sschaef:
If I would have known that it is so easy to implement this, I would have
done it earlier. I don't know if anyone else needs this, but I
definitely need it. When I explicitly create a wildcard import, I do not
want that it gets expanded, otherwise I wouldn't have created a wildcard
import in the first place. The other option, which preserves everything,
is not what I want to use, since multi line imports definitely should be
expanded if they exist somewhere.

Fixes #1002574


wpopielarski:
LGTM

PR #1021, opened 2015-10-23
Fix NPE in SbtInputs

sschaef:
An user found another NPE in SbtInputs. It is solved in the same way as
the previous NPE was solved. If a path is invalid an
IllegalStateException is thrown, which is a little bit better than a
NPE, which doesn't give any hints about the problem.

Fixes #1002577


wpopielarski:
LGTM

PR #1022, opened 2015-10-26
Additional filtering for Step Out Message

wpopielarski:
Problem: when `receive` method is wrapped by other `receive` method
the Step Out processing stops in first found `applyOrElse` method.
Example:
```
def receive = LoggingReceive {
  case msg => foo
}
```
so Step Out stops in implementation of LoggingReceive not in function
shown above.

Perscription:
In Launch Configuration in Scala Debug Tab filter for Step Out Message
functionality has been added. The `applyOrElse` from listed package/class
names are ignored if found (blacklisted).

fix #1002575

### DIFF
--- a/org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/async/StepMessageOut.scala
+++ b/org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/async/StepMessageOut.scala
@@ -31,10 +31,10 @@ import com.sun.jdi.request.StepRequest
 case class StepMessageOut(debugTarget: ScalaDebugTarget, thread: ScalaThread) extends HasLogger {
   import org.scalaide.debug.internal.launching.ScalaDebuggerConfiguration._
   import scala.collection.JavaConverters._
-  private val StepMessageOut = "StepMessageOut" + UUID.randomUUID.toString
-  private val IsSearchingReceiveMethodKey = "search"
-  private val StackTraceDepth = "StackTraceDepth"
-  private val StartLineNumber = "StartLineNumber"
+  private val StepMessageOut = "org.scala-ide.debug.async.step-out.StepMessageOut" + UUID.randomUUID.toString
+  private val IsSearchingReceiveMethodKey = "org.scala-ide.debug.async.step-out.Search"
+  private val StackTraceDepth = "org.scala-ide.debug.async.step-out.StackTraceDepth"
+  private val StartLineNumber = "org.scala-ide.debug.async.step-out.StartLineNumber"
   private var watchedMessage: Option[ObjectReference] = None
   private var watchedActor: String = _
   val programSends = List(
@@ -62,23 +62,18 @@ case class StepMessageOut(debugTarget: ScalaDebugTarget, thread: ScalaThread) ex
   private[async] def findThisType(stepEvent: StepEvent): String = {
     val loc = stepEvent.location()
     val declType = loc.declaringType().name()
-    Try {
-      val thisObject = stepEvent.thread().frame(0).thisObject()
-      if (thisObject ne null)
-        thisObject.referenceType().name()
-      else
-        declType
-    }.recover {
-      case e: Throwable =>
-        declType
-    }.get
+    val thisObject = stepEvent.thread().frame(0).thisObject()
+    if (thisObject ne null)
+      thisObject.referenceType().name()
+    else
+      declType
   }
 
   object subordinate extends BaseDebuggerActor {
     import scala.collection.JavaConverters._
     private def isReceiveHandler(stepEvent: StepEvent): Boolean = {
       val thisType = findThisType(stepEvent)
-      stepEvent.location.method.name.contains("applyOrElse") &&
+      "applyOrElse" == stepEvent.location.method.name &&
         !stepOutNotStopInClasses.exists { thisType.startsWith }
     }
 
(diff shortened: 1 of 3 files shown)


org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/async/StepMessageOut.scala
```diff
@@ -50,38 +59,57 @@ case class StepMessageOut(debugTarget: ScalaDebugTarget, thread: ScalaThread) ex
     thread.resumeFromScala(DebugEvent.CLIENT_REQUEST)
   }
 
+  private[async] def findThisType(stepEvent: StepEvent): String = {
+    val loc = stepEvent.location()
+    val declType = loc.declaringType().name()
+    Try {
```
sschaef: You could just use try-catch here, couldn't you? Also why do you just drop the throwable, how can it happen?

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/async/StepMessageOut.scala
```diff
@@ -50,38 +59,57 @@ case class StepMessageOut(debugTarget: ScalaDebugTarget, thread: ScalaThread) ex
...
+    }.get
+  }
+
   object subordinate extends BaseDebuggerActor {
     import scala.collection.JavaConverters._
-    private def isReceiveHandler(loc: com.sun.jdi.Location): Boolean =
-      loc.method().name().contains("applyOrElse")
+    private def isReceiveHandler(stepEvent: StepEvent): Boolean = {
+      val thisType = findThisType(stepEvent)
+      stepEvent.location.method.name.contains("applyOrElse") &&
```
sschaef: Why `contains` and not `==`

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/launching/ScalaDebuggerTab.scala
```diff
@@ -44,19 +65,27 @@ private class GeneralSettingsBlock(val tab: ScalaDebuggerTab) {
...
-  def performApply(configuration: ILaunchConfigurationWorkingCopy): Unit =
+  def performApply(configuration: ILaunchConfigurationWorkingCopy): Unit = {
     if (globalAsyncDebuggerSwitchButton.getSelection()) {
       configuration.setAttribute(LaunchWithAsyncDebugger, true)
     } else {
       configuration.removeAttribute(LaunchWithAsyncDebugger)
     }
+    import scala.collection.JavaConverters._
+    configuration.setAttribute(StepOutExcludePkgsOrClasses,
+      stepOutExcludePackages.getText.split(DataDelimiter).filterNot { _.isEmpty }.toList.asJava)
```
sschaef: I think there is `nonEmpty`

wpopielarski: not sure, this is method on String, let me check and if there is `nonEmpty` I replace it (is more natural)

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/async/StepMessageOut.scala
wpopielarski: well, I think that there was something like trial and error in the past. So looks like 'apply' was considered too... Now looks like torpidity. I try it and replace

org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/async/StepMessageOut.scala
wpopielarski: Try is so functional... 
Sometimes JDI is not too responsive and for example invalidates the stack frame/trace on debugging machine (I'm not sure if my explanation is clear). So here we could have some strategies to take:
1. allow exception to be thrown and debugging process is stopped
2. try another call with JDI and if succeeded then continue debugging otherwise rethrow and break
3. drop to some default and allow debugging, the method to stop is not found at most.
The problem for me is just to assess what fraction of JDI calls ends up with exception...

sschaef:
I'm not sure if it is wise to put the configuration for the new functionality in the run configuration tab. I think it would be better to put it in Async Debugger preferences and make it possible to overwrite it by project configuration. Otherwise, I'm afraid people are not going to find it.

The same actually applies for the Async Debugger itself. We can globally enable/disable it in the preferences and overwrite behavior in project configuration. WDYT?

wpopielarski:
At last I'm back after Expression Evaluator work. Sorry for late opinion. So I feel that settings of filtering at Launch Config level is quite vindicated. In one test user can want to stop in some implementation in second doesn't. Setting it global is a surplus for me. I'm aware that user likely doesn't stop in applyOrElse of PartialFunction and usually in any extension of it in akka packages but especially the second one might not be so often needed as I think. If we are not ok with this configuration maybe we should open a ticket in assembla?

sschaef:
For me easiness of usage counts higher than the lost possibility of having multiple run configurations with different behavior (which seems to me quite unlikely for a user to have). But I also don't have any statistics on that so it is just the feeling. I guess if we document all the stuff in the run configuration tab, than at least we can point to it in the release notes and when someone asks on the mailing lists.

wpopielarski:
retest it please

wpopielarski:
retest this please

sschaef:
LGTM

PR #1023, opened 2015-10-29
Fix transitive dependencies in ScalaProject

romanowski:
Previously transitive dependencies from ScalaProject were limited to dependencies of dependencies. Currenty deps are computed recursively.

### DIFF
--- a/org.scala-ide.sdt.core.tests/src/org/scalaide/core/sbtbuilder/ProjectDependenciesTest.scala
+++ b/org.scala-ide.sdt.core.tests/src/org/scalaide/core/sbtbuilder/ProjectDependenciesTest.scala
@@ -14,40 +14,87 @@ import org.eclipse.jdt.core.IPackageFragment
 import org.scalaide.util.internal.SettingConverterUtil
 import org.scalaide.ui.internal.preferences.ScalaPluginSettings
 import org.scalaide.core.IScalaPlugin
+import org.eclipse.core.resources.IProject
 
 class ProjectDependenciesTest {
 
   import SDTTestUtils._
 
+  private implicit class TestableProject(from: IScalaProject) {
+    def dependsOnAndExports(dep: IScalaProject, exported: Boolean = true): Unit =
+      addToClasspath(from, JavaCore.newProjectEntry(dep.underlying.getFullPath, exported))
+
+    def onlyDependsOn(dep: IScalaProject) = dependsOnAndExports(dep, false)
+
+    def shouldDependOn(msg: String, deps: IScalaProject*): Unit = {
+      val expected = deps.map(_.underlying).sortBy(_.getName)
+      val computed = from.transitiveDependencies.sortBy(_.getName)
+      Assert.assertEquals(s"${msg.capitalize} for ${from.underlying.getName}", expected, computed)
+    }
+  }
+
+
+  @Test def transitive_dependencies_more_complicated_tree(): Unit = {
+    val allProj @ Seq(prjA, prjB, prjC, prjD, prjE, prjF, prjG) = createProjects("A", "B", "C", "D", "E", "F", "G")
+    try {
+      prjB dependsOnAndExports prjA
+
+      prjD dependsOnAndExports prjC
+
+      prjE onlyDependsOn prjB
+
+      prjE dependsOnAndExports prjD
+
+      prjF dependsOnAndExports prjA
+      prjF dependsOnAndExports prjE
+
+      prjG dependsOnAndExports prjE
+      prjG dependsOnAndExports prjC
+
+
+      prjB.shouldDependOn("Only A", prjA)
+      prjD.shouldDependOn("Only C", prjC)
+      prjE.shouldDependOn("A, B, C,D", prjB, prjD, prjA, prjC)
+      prjG.shouldDependOn("C, D, E - C should not be added twice, B and A should be excluded", prjC, prjE, prjD)
+
+    } finally {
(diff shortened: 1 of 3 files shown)


sschaef:
Scalastyle error:
> error file=/localhome/jenkinside/workspace/ghprb-scala-ide-multi-conf/eclipse_platform/eclipse-luna/scala-ide/org.scala-ide.sdt.core/src/org/scalaide/core/internal/project/ScalaProject.scala message=Whitespace at end of line line=229 column=12

org.scala-ide.sdt.core.tests/src/org/scalaide/core/sbtbuilder/ProjectDependenciesTest.scala
```diff
@@ -14,40 +14,87 @@ import org.eclipse.jdt.core.IPackageFragment
...
 
   import SDTTestUtils._
 
+  private implicit class TestableProject(from: IScalaProject) {
+    def dependsOnAndExports(dep: IScalaProject, exported: Boolean = true): Unit =
+      addToClasspath(from, JavaCore.newProjectEntry(dep.underlying.getFullPath, exported))
+
+    def onlyDependsOn(dep: IScalaProject) = dependsOnAndExports(dep, false)
+
+    def shouldDependsOn(msg: String, deps: IScalaProject*): Unit = {
```
sschaef: `s/Depends/Depend/`

org.scala-ide.sdt.core.tests/src/org/scalaide/core/sbtbuilder/ProjectDependenciesTest.scala
```diff
@@ -14,40 +14,87 @@ import org.eclipse.jdt.core.IPackageFragment
...
+      prjG dependsOnAndExports prjE
+      prjG dependsOnAndExports prjC
+
+
+      prjB.shouldDependsOn("Only A", prjA)
+      prjD.shouldDependsOn("Only C", prjC)
+      prjE.shouldDependsOn("A, B, C,D", prjB, prjD, prjA, prjC)
+      prjG.shouldDependsOn("C, D, E - C should not be added twice, B and A should be excluded", prjC, prjE, prjD)
+
+    } finally {
```
sschaef: You could move the `deleteProjects` to a method annotation with `@After` and make `allProj` a class value.

org.scala-ide.sdt.core/src/org/scalaide/core/internal/project/ScalaProject.scala
```diff
@@ -131,9 +132,37 @@ object ScalaProject {
...
+    if (projectsToTest.isEmpty)
+      exportedProjects
+    else {
+      exportedDependenciesForProjects(
+        projectsToTest.flatMap(dependenciesForProject).map(projectFromPath),
+        exportedProjects union projectsToTest)
+    }
+  }
+
+  def projectFromPath(project: IPath): IProject = EclipseUtils.workspaceRoot.getProject(project.toString)
```
sschaef: I think it is better to move this to `FileUtils`

sschaef:
Would you please open a ticket in Assembla with a small description what the problem is and what needs to be improved and mark it as fixed by your commit?

org.scala-ide.sdt.core.tests/src/org/scalaide/core/sbtbuilder/ProjectDependenciesTest.scala
romanowski: I follow convention in this file. I this case I need to create a var and I don't like that.
But I can refactor this you really like ;)

sschaef: No, it is ok. I was just saying.

romanowski:
Ticket:
https://www.assembla.com/spaces/scala-ide/support/tickets/1002581-scala-ide-incremental-compiler-sees-only-depenecies-of-depenecies

sschaef:
The commit message also needs to say `Fix #1002581` at the end.

sschaef:
Even with that patch, I can't that up and running in Eclipse. I did: `mkdir {a,b,c,d}/src/main/scala`. Then I added Scala files as described in the ticket. I typed `sbt eclipse`, imported the projects in Eclipse and got an error message in `a`because the definition in `d` was not found. Anything else one has to consider by creating such a deep project hierarchy?

sschaef:
Ok, I'm stupid. I wrote broken code. Now everything compiles. The thing is, it even compiles without the patch. I also can do changes and they are propagated through all projects the moments when I start building. PC doesn't show errors before a build but that is normal I would say.

romanowski:
Hierarchy this deep  is common in really big projects.

The problem is that even when I change code in D that will make A not compiling (add additional parameter do D.d() method) incremental compilation shows no errors (but it should).
Clean compile fixing it.

sschaef:
I still can't see any compilation errors propagated to the most outer project. When I change D, the log only shows:
```
2015-10-30 18:30:25,265  INFO [Worker-5] - ScalaBuilder - Building project d
...
2015-10-30 18:30:26,750 DEBUG [Worker-5] - ScalaProject - [d] Reset PC of referring project c
```
I still have to trigger a recompilation in A to get a recompilation of C, B and A.

sschaef:
It turned out in a hangouts meeting that I didn't understand what I was doing. LGTM then.

PR #1024, opened 2015-10-30
Do not apply cursor position translation twice

sschaef:
`modelOffset2WidgetOffset` translates positions from the document to
positions of the shown widget, which may show folded regions. In these
cases the cursor was set to the wrong position. The problem was that the
translations happened twice. Once in the changed code and once in
`setSelectedRange`. By removing one translation, the wrong cursor
positions are sent to oblivion.

Fixes #1002579


dragos:
LGTM

PR #1025, opened 2015-11-02
Expr eval enhancements

wpopielarski:
Set of commits which amend usage of EE a little bit:
- problem of evaluation in nested classes is solved
- switch for importing of packages from current file is added to weaken a problem of toolbox with nested packages
- test for nested scala object has been added for assembla ticket #1002582

### DIFF
--- a/org.scala-ide.sdt.debug.expression/src/org/scalaide/debug/internal/expression/ExpressionEvaluator.scala
+++ b/org.scala-ide.sdt.debug.expression/src/org/scalaide/debug/internal/expression/ExpressionEvaluator.scala
@@ -6,11 +6,12 @@ package org.scalaide.debug.internal.expression
 import scala.reflect.runtime.universe
 import scala.tools.reflect.ToolBox
 import scala.util.Try
+
 import org.scalaide.debug.internal.expression.context.JdiContext
 import org.scalaide.debug.internal.expression.context.VariableContext
 import org.scalaide.debug.internal.expression.proxies.JdiProxy
 import org.scalaide.debug.internal.expression.proxies.phases._
-import org.scalaide.debug.internal.preferences.ExpressionEvaluatorPreferences.isAddImportsFromCurrentFile
+import org.scalaide.debug.internal.preferences.ExpressionEvaluatorPreferences.shouldAddImportsFromCurrentFile
 import org.scalaide.logging.HasLogger
 
 object ExpressionEvaluator {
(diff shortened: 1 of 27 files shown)


org.scala-ide.sdt.debug.expression/src/org/scalaide/debug/internal/expression/ExpressionEvaluator.scala
```diff
@@ -6,11 +6,11 @@ package org.scalaide.debug.internal.expression
 import scala.reflect.runtime.universe
 import scala.tools.reflect.ToolBox
 import scala.util.Try
-
```
Kwestor: Bring this line back! ;)

org.scala-ide.sdt.debug.expression/src/org/scalaide/debug/internal/preferences/ExpressionEvaluatorPreferences.scala
```diff
@@ -46,4 +46,7 @@ object ExpressionEvaluatorPreferences {
 
   def shouldRefreshVariablesViewAfterEvaluation: Boolean =
     preferenceStore.getBoolean(RefreshVariablesViewAfterEvaluation)
+
+  def isAddImportsFromCurrentFile: Boolean =
```
Kwestor: `shouldAddImportsFromCurrentFile`?

Kwestor:
LGTM

sschaef:
I don't think it is a good design decision to add the new option to the preferences. Users would not understand what it means even if it is documented. And even if they understand it, no one wants to care about such internal details. It justs adds complexity to the usage of the IDE and to its internals. We should fix the problem at the root. The other commits with the tests are ok, but I don't like the one with the new preference option.

wpopielarski:
Looks like there is needed deeper reverie about add imports. As I see it
there are 3 options by now:
- let it be as it is so in many cases the evaluation won't be possible
(nested packages in toolbox limitation)
- utterly skip 'AddImports' phase in EE context building so user needs to
use fully qualified names or add import statement by hand
- stick to switch and allow user to decide
These solutions are by now. I open a ticket in assembla to recognize and
implement workaround for toolbox problem
WDYT?


2015-11-03 23:08 GMT+01:00 Simon Schäfer <notifications@github.com>:

wpopielarski:
UPDATE:
in case of nested packages fully qualified names work only, so adding imports by hand makes same problems like adding them with phase

sschaef:
I think it would be best if EE just analyzes what needs to be imported and then creates the correct import. It wouldn't solve the problem with Toolbox, but at least it wouldn't create wrong code anymore.

wpopielarski:
Could you describe it more detailed? I opened a ticket #1002584. Asking,
because I don't know what's your solution idea.

2015-11-04 12:00 GMT+01:00 Simon Schäfer <notifications@github.com>:

sschaef:
retest this please

sschaef:
LGTM

PR #1030, opened 2015-11-15
Fix AIOOBE in outline view for 0-arity functions

sschaef:
Fix [#1002592](https://www.assembla.com/spaces/scala-ide/tickets/1002592-outline-view-not-working-because-of-aioobe)

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/outline/ModelBuilder.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/outline/ModelBuilder.scala
@@ -41,11 +41,18 @@ object ModelBuilder extends HasLogger {
     }
 
     def updateTree(parent: ContainerNode, t: Tree): Unit = {
+      def renderArgs[A](sb: StringBuilder, args: List[A])(f: A => Unit) = {
+        args match {
+          case head :: tail =>
+            f(head)
+            tail.foreach { a => sb.append(", "); f(a) }
+          case _ =>
+        }
+      }
+
       def renderTuple(sb: StringBuilder, args: List[Tree]) = {
         sb.append("(")
-        args.foreach(a => { renderType(sb, a); sb.append(", ") })
-        if (!args.isEmpty)
-          sb.setLength(sb.length - 2)
+        renderArgs(sb, args)(renderType(sb, _))
         sb.append(")")
       }
 
@@ -56,8 +63,7 @@ object ModelBuilder extends HasLogger {
           renderType(sb, args.last)
         } else {
           sb.append("(")
-          args.dropRight(1).foreach(x => { renderType(sb, x); sb.append(", ") })
-          sb.setLength(math.max(1, sb.length - 2))
+          renderArgs(sb, args.dropRight(1))(renderType(sb, _))
           sb.append(") => ")
           renderType(sb, args.last)
         }
@@ -104,8 +110,7 @@ object ModelBuilder extends HasLogger {
             sb.append(showName(name, tt))
             if (!args.isEmpty) {
               sb.append("[")
-              args.foreach(a => { renderType(sb, a, List(), false); sb.append(", ") })
-              sb.setLength(sb.length - 2)
+              renderArgs(sb, args)(renderType(sb, _, Nil, needParenthesis = false))
               sb.append("]")
             }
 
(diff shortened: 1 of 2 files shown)


org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/outline/ModelBuilder.scala
```diff
@@ -57,7 +57,7 @@ object ModelBuilder extends HasLogger {
         } else {
           sb.append("(")
           args.dropRight(1).foreach(x => { renderType(sb, x); sb.append(", ") })
-          sb.setLength(sb.length - 2)
+          sb.setLength(math.max(1, sb.length - 2))
```
wpopielarski: I don't want to be a stickler but what do `1` and `sb.length - 2` mean?

sschaef: 1 is the length of "(" and 2 is the length of ", ", which is added once too often by the previous line. I didn't know how to better fix the code without refactoring half of it.

misto: You could write "(".length and ", ".length instead, that would make it easier to understand.

sschaef: I just wanted to follow the style of the rest of the code. ;) But you are right, will do.

sschaef: Fixed

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/outline/ModelBuilder.scala
```diff
@@ -41,11 +41,16 @@ object ModelBuilder extends HasLogger {
     }
 
     def updateTree(parent: ContainerNode, t: Tree): Unit = {
+      def renderArgs[A](sb: StringBuilder, args: List[A])(f: A => Unit) = {
+        if (args.nonEmpty) {
+          f(args.head)
+          args.drop(1).foreach { a => sb.append(", "); f(a) }
```
wpopielarski: why do you have to drop first arg?

sschaef: Because it is already processed in the line above.

wpopielarski: I'm not so perceivable :(. Actually this code looks little strange. Now I see `f` should use same instance of `sb` but is not explicitly expressed. I'm aware you are constrained by existing implementation....

dragos: I agree with @wpopielarski that this could be simpler by replacing everything with something along the lines of

```
args.map(renderType(sb, _)).mkString(", ")
```

dragos: or, alternatively, by using pattern matching:

```
args match { 
  case head :: tail => ...
  case Nil => 
}
```

sschaef: I moved to the pattern match now. For me that is good enough.

sschaef:
@dragos Do you understand this error message:

```
ouldn't read comment file
java.io.FileNotFoundException: File 'scala-ide/commentinfo.md' does not exist
	at org.apache.commons.io.FileUtils.openInputStream(FileUtils.java:299)
	at org.apache.commons.io.FileUtils.readFileToString(FileUtils.java:1711)
	at org.apache.commons.io.FileUtils.readFileToString(FileUtils.java:1748)
	at org.jenkinsci.plugins.ghprb.extensions.comments.GhprbCommentFile.postBuildComment(GhprbCommentFile.java:45)
```

dragos:
Yeah, I was trying to customize the comment message with a link to the update site, it's in the job configuration. You can specify the name of a file, and that gets included in the message. I'm trying to generate it in the job config, but it seems it's not done, I have no idea why..

PR #1031, opened 2015-11-16
Show ctors as `this` and not as `<init>` it outline view

sschaef:
Fix [#1002588](https://www.assembla.com/spaces/scala-ide/tickets/1002588-class-constructors-are-shown-as--lt-init-gt--in-outline-view)

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/outline/ModelBuilder.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/outline/ModelBuilder.scala
@@ -153,10 +153,15 @@ object ModelBuilder extends HasLogger {
       }
 
       def showName(name: Name, t: Tree) = {
-        if (src.content(t.pos.point) == '`')
-          "`" + name.decoded + "`"
-        else
-          name.decoded.split("\\.").reverse.head
+        t match {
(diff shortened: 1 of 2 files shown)


wpopielarski:
LGTM

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/outline/ModelBuilder.scala
```diff
@@ -153,10 +153,15 @@ object ModelBuilder extends HasLogger {
       }
 
       def showName(name: Name, t: Tree) = {
-        if (src.content(t.pos.point) == '`')
-          "`" + name.decoded + "`"
-        else
-          name.decoded.split("\\.").reverse.head
+        t match {
+          case t: DefDef if t.name.toString == "<init>" =>
```
dragos: nitpick: you can compare `t.name` with `nme.CONSTRUCTOR`. It's two Int comparisons, and you save both an allocation and a linear comparison via Strings :)

Kwestor: `nme` is generally useful when matching trees :)

sschaef: Fixed

PR #1033, opened 2015-11-17
Fix failing accessibility tests

sschaef:
The tests failed because the accessor was shown in the display string.
Removing the accessor was not as easy as I hoped. The problem was that
the private[x] keyword is not represented as a flag. The only way to
know that a symbol is private for a package is by checking if it has no
other visibility.  This meant that the already existing flag mask
couldn't be used, since the private keyword just couldn't be masked out.
How stupid.

I fixed it by adding another parameter `showAccessor`, which needs to be
set to false when the accessor shouldn't be shown. For public, protected
and normal private visibility it is also possible to mask the flags with
the `flagMask` parameter. Just not for private[x].

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/core/compiler/DeclarationPrinter.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/core/compiler/DeclarationPrinter.scala
@@ -39,7 +39,7 @@ abstract class DeclarationPrinter extends HasLogger {
    *
    *  @note Should be run on the presentation compiler thread (inside askOption).
    */
-  def defString(sym: Symbol, flagMask: Long = Flags.ExplicitFlags, showKind: Boolean = true)(seenAs: Type = sym.rawInfo): String = {
+  def defString(sym: Symbol, flagMask: Long = Flags.ExplicitFlags, showKind: Boolean = true, showAccessor: Boolean = true)(seenAs: Type = sym.rawInfo): String = {
 
     def infoString(tp: Type): String = {
       def isStructuralThisType = (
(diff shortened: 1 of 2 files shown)


org.scala-ide.sdt.core/src/org/scalaide/core/compiler/DeclarationPrinter.scala
```diff
@@ -67,12 +67,10 @@ abstract class DeclarationPrinter extends HasLogger {
     /* Inspired from Symbol and HasFlags, but take into account the mask access modifiers.
      * The original would always display access modifiers, even when the mask is `0L`.
      */
```
wpopielarski: Not sure if this description is valid now

org.scala-ide.sdt.core/src/org/scalaide/core/compiler/DeclarationPrinter.scala
```diff
@@ -39,7 +39,7 @@ abstract class DeclarationPrinter extends HasLogger {
    *
    *  @note Should be run on the presentation compiler thread (inside askOption).
    */
-  def defString(sym: Symbol, flagMask: Long = Flags.ExplicitFlags, showKind: Boolean = true)(seenAs: Type = sym.rawInfo): String = {
+  def defString(sym: Symbol, flagMask: Long = Flags.ExplicitFlags, showKind: Boolean = true, showAccessor: Boolean = true)(seenAs: Type = sym.rawInfo): String = {
```
wpopielarski: I'm just thinking but instead of `showAccessor` flag maybe inner function could be used like:
`def showAccessor = flagMask != 0L && (flagMask & AccessFlags) == 0  ?

org.scala-ide.sdt.core/src/org/scalaide/core/compiler/DeclarationPrinter.scala
```diff
@@ -67,12 +67,10 @@ abstract class DeclarationPrinter extends HasLogger {
...
      */
-    def flagString(sym: Symbol, basis: Long): String = {
-      import sym.{ accessString => _, _ }
+    def flagString(sym: Symbol): String = {
       import Flags._
-      val access = accessString(sym)
-
-      val nonAccess = flagBitsToString(basis & ~AccessFlags)
+      val access = if (showAccessor) accessString(sym) else ""
+      val nonAccess = sym.flagBitsToString(sym.flags & flagMask & ~AccessFlags)
```
wpopielarski: generally your refactoring is pretty ok, but I would implement `flagString` as
```
    def flagString(sym: Symbol): String = {
      import Flags._
      val access = if ((flagMask & AccessFlags | flagMask) != 0) accessString(sym) else ""
      val nonAccess = sym.flagBitsToString(sym.flags & flagMask & ~AccessFlags)

      if (access == "") nonAccess
      else if (nonAccess == "") access
      else nonAccess + " " + access
    }
```

and then you don't need `showAccessor` flag. Is it ok (this my change passes tests but I don't know if there are other scenarios)?

sschaef:
Superseded by #1035

PR #1035, opened 2015-11-18
Alternative #1033-fix failing accessibility tests

wpopielarski:
Simon, it is just proposal, you can take it or supersede by yours.


sschaef:
Ok, that one is simpler than mine. LGTM

PR #1036, opened 2015-11-19
Outline view folds import nodes by default

andrey-ilinykh:
This feature is controlled by a preference Scala/Editor/Outline

Fix #1002595


org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/outline/OutlineHelper.scala
```diff
@@ -0,0 +1,13 @@
+package org.scalaide.ui.internal.editor.outline
+
+import org.eclipse.jface.viewers.TreeViewer
+
+object OutlineHelper {
+  def foldImportNodes(viewer: TreeViewer, inout: Any): Unit = {
```
sschaef: `s/inout/input/`

sschaef:
LGTM

PR #1037, opened 2015-11-20
Refresh analysis store after build

wpopielarski:
Scenario comes from one of customer who updates .cache-* by external
to Scala IDE tool (strictly by gradle build script but it does not
matter).
Scenario:
- open project in Scala IDE
- mimic external tool run by editing any of .cache-* file and change
  its content
- put some change to any scala file and build a project
- try Eclipse 'search'
Without the change error popup window is seen with information about
a loss of sync of .cache-* file (so asking about using F5 on project).
The suggested improvement assumes to refresh .cache-* files after
project build.

REMARK: No test unit, looks too hard to write, any suggestion to create welcomed


Kwestor:
LGTM

sschaef:
LGTM

PR #1039, opened 2015-11-23
Don’t create a Scala reconciler for Scala files that are not part of a

dragos:
source folder.

This makes it easier to use the Scala editor on files like Sbt, or simple script examples (by not filling them up with error markers).

Ref #1002415


sschaef:
LGTM

PR #1041, opened 2015-11-30
Statistics tracker improvements

sschaef:
See commit messages for more details.

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/core/internal/statistics/Statistics.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/core/internal/statistics/Statistics.scala
@@ -7,7 +7,9 @@ import scala.collection.JavaConverters._
 
 import org.scalaide.core.ScalaIdeDataStore
 import org.scalaide.core.internal.ScalaPlugin
+import org.scalaide.logging.HasLogger
 
+import com.cedarsoftware.util.io.JsonIoException
 import com.cedarsoftware.util.io.JsonReader
 import com.cedarsoftware.util.io.JsonWriter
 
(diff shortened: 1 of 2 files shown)


dragos:
Incrementing an in-memory counter shouldn't take too much time. In fact, it might be way slower now, since the overhead of a Job is quite significant. My remark previously was about loading the Statistics file, which involves disk access, but if that is done only once, on startup, I think it's all fine.

org.scala-ide.sdt.core/src/org/scalaide/core/internal/statistics/Statistics.scala
```diff
@@ -16,9 +21,12 @@ import Features._
 /**
  * Contains all definitions that belong to the statistics tracker.
  */
-class Statistics {
+class Statistics extends AnyRef with HasLogger {
```
wpopielarski: just curious about extending AnyRef, any hidden scala feature?

wpopielarski:
actually LGTM

org.scala-ide.sdt.core/src/org/scalaide/core/internal/statistics/Statistics.scala
sschaef: No hidden feature. I just don't like to use `extend` with a trait. Unnecessary thoughts...

sschaef:
On 12/01/2015 10:36 AM, Iulian Dragos wrote:
>
> Incrementing an in-memory counter shouldn't take too much time. In 
> fact, it might be way slower now, since the overhead of a Job is quite 
> significant. My remark previously was about loading the Statistics 
> file, which involves disk access, but if that is done only once, on 
> startup, I think it's all fine.
>
Indeed, now after you said it I'm going to remove that commit. I thought 
that because we not only increment but also write back the chances to 
disk, it could take some time. But some benchmarks showed that is still 
doesn't even need 1ms on a SSD. Don't know how huge the slowdown is for 
a HDD but probably negligible.

PR #1042, opened 2015-11-30
Makes scala installation check less vulnerable

wpopielarski:


sschaef:
I want to use the possibility and criticize your commit message once again. ;) You could have mentioned why the change is needed and if it is only a copy of Iulians message on gitter, which already summarizes the problem quite well: https://gitter.im/scala-ide/scala-ide?at=565c2515c3d575114e6cb2d1

sschaef:
Fight to the end, I'm sure you will win!

wpopielarski:
my problem is that I cannot do it locally so I'm blindfold a little

dragos:
I tried to have a look at the PR, but it's a bit too complicated for a cursory review. I wonder if this won't make it harder to undo the change, once `scala-actors` starts using the right version string.

dragos:
In the mean time, and given that tests pass :) I would say LGTM

wpopielarski:
just `ScalaInstallationTest` is little tricky. There is the one change only in `ScalaInstallation`, just in method `private def findBundle(bundleId: String, version: Version): Option[Bundle]`. The rest of changes come from formatting + organize imports

sschaef:
formatting should be in its own commit actually. It is so annoying to find the actual semantic changes within all the changes. The very first version in contrast was clear and understandable (and didn't work but that is beside the point).

sschaef:
As discussed in the meeting, it looks ugly but LGTM ;)

PR #1044, opened 2015-12-12
Make reconciliation time configurable

sschaef:
Users can now select a different time through the Scala preference page.

Fixes #1002612

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/ui/internal/preferences/ScalaPreferences.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/ui/internal/preferences/ScalaPreferences.scala
@@ -4,15 +4,22 @@ import java.io.File
 
 import org.eclipse.core.runtime.preferences.AbstractPreferenceInitializer
 import org.eclipse.jface.preference.FieldEditorPreferencePage
+import org.eclipse.jface.preference.IntegerFieldEditor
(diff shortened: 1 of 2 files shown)


org.scala-ide.sdt.core/src/org/scalaide/ui/internal/preferences/ScalaPreferences.scala
```diff
@@ -27,14 +34,26 @@ class ScalaPreferences extends FieldEditorPreferencePage with IWorkbenchPreferen
...
 
-  def init(workbench: IWorkbench): Unit = ()
+  override def init(workbench: IWorkbench): Unit = ()
+
+  override def performOk() = {
+    super.performOk()
+    val store = IScalaPlugin().getPreferenceStore
+    store.setValue(ScalaIdeDataStore.DataStoreId, dataStoreField.getStringValue)
+    store.setValue(ScalaPreferences.ReconcilerDelayId, reconcilerField.getStringValue)
+    true
```
wpopielarski: shouldn't return with `super.performOk() && true` or even `super.performOk()`? Just curios.

sschaef: I just looked it up, `super.performOk()` returns always `true` too. No need to change then.

wpopielarski:
LGTM

PR #1045, opened 2015-12-12
Add auto edits for better backtick handling

sschaef:
Fixes #1002610

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/extensions/autoedits/CloseBacktick.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/extensions/autoedits/CloseBacktick.scala
@@ -1,0 +1,51 @@
+package org.scalaide.extensions
+package autoedits
+
+import org.eclipse.jface.text.IDocument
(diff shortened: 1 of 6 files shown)


org.scala-ide.sdt.core/src/org/scalaide/extensions/autoedits/CloseBacktick.scala
```diff
@@ -0,0 +1,51 @@
...
+
+trait CloseBacktick extends AutoEdit {
+
+  override def setting = CloseBacktickSetting
+
+  override def perform() = {
+    def ch(i: Int, c: Char) = {
+      val o = textChange.start + i
+      o >= 0 && o < document.length && document(o) == c
+    }
```
wpopielarski: looks like test for character `c` at position `start + i` but used locally with `-1` and c ```. So maybe easier would rename it to `isStartWithBacktick` ?

sschaef: done

wpopielarski:
LGTM

PR #1048, opened 2015-12-17
Hide quick fix entry from context menu

sschaef:
The quick fix entry is contributed by JDT and is therefore broken since
it invokes only the JDT quick assists.

Fixes #1002523

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ScalaSourceFileEditor.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ScalaSourceFileEditor.scala
@@ -222,7 +222,12 @@ class ScalaSourceFileEditor
     }
 
     def fixBrokenGroups(): Unit = {
-      findJdtSourceMenuManager(menu.getItems) foreach { mm =>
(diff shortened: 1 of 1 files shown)


org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/ScalaSourceFileEditor.scala
```diff
@@ -222,7 +222,12 @@ class ScalaSourceFileEditor
     }
 
     def fixBrokenGroups(): Unit = {
-      findJdtSourceMenuManager(menu.getItems) foreach { mm =>
+      val items = menu.getItems
+
+      // hide quick fix entry from menu since it invokes JDT quick fixes
+      items.find(_.getId == "QuickAssist").map(_.setVisible(false))
```
wpopielarski: `foreach` ?

sschaef: good catch

sschaef:
retest this please

wpopielarski:
LGTM

PR #1049, opened 2015-12-18
Cleanups

sschaef:
See commit messages for changes.


sschaef:
Merge without review. Blocks me otherwise.

PR #1050, opened 2015-12-27
Add update-site for scala-refactoring

sschaef:
Right now, the scala-refactoring repo contains an update-site, which is
used by the Scala IDE build. Since the update-site just clutters up the
scala-refactoring repo, the idea was to move the update-site to the
scala-ide repo. This way the scala-refactoring repo is easier to
understand. The scala-ide repo gets more complicated but we can hide all
of these artificial update sites in a sub directory instead of keeping
them in the root directory as it is done right now.

Another change that was required to make this up and running can be
found in the nightlies of scala-refactoring. So far a P2 update site was
published by our build server. This has changed in so far that a nightly
of the scala-refactoring library is published, which can also be
consumed by other people.

### DIFF
--- a/org.scala-ide.refactoring/pom.xml
+++ b/org.scala-ide.refactoring/pom.xml
@@ -1,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.scala-ide</groupId>
+    <artifactId>org.scala-ide.p2-toolchain</artifactId>
+    <version>4.4.0-SNAPSHOT</version>
+    <relativePath>../org.scala-ide.p2-toolchain/pom.xml</relativePath>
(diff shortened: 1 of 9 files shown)


sschaef:
The Jenkins job that publishes scala-refactoring nightlies is https://jenkins.scala-ide.org:8496/jenkins/view/Parameterized%20Nightly%20Jobs/job/parameterized-refactoring/

org.scala-ide.refactoring/pom.xml
```diff
@@ -0,0 +1,42 @@
...
+    <groupId>org.scala-ide</groupId>
+    <artifactId>org.scala-ide.p2-toolchain</artifactId>
+    <version>4.4.0-SNAPSHOT</version>
+    <relativePath>../org.scala-ide.p2-toolchain/pom.xml</relativePath>
+  </parent>
+
+  <artifactId>org.scala-ide.refactoring</artifactId>
+  <description>The local p2 repo for scala-refactoring</description>
+  <packaging>pom</packaging>
+  <version>4.4.0-SNAPSHOT</version>
```
misto: Couldn't in be confusing if scala-refactoring has two version numbers?

sschaef: This version is the one of the P2 repo, it is only used internally in our build process. I don't think users will ever see it. Keeping this version in sync with the version of the IDE for simplicity reasons is probably the best.

misto: Ok, that makes sense! Looks good to me.

PR #1051, opened 2015-12-28
Adds missing TRACE debug level

wpopielarski:


sschaef:
There should be exhaustive warnings for enums. LGTM

PR #1052, opened 2015-12-28
Add update-site for scalariform

sschaef:
This change works in the same way as the change for the
scala-refactoring update-site in commit b6ed45040c1.


sschaef:
wrong merge branch

PR #1053, opened 2015-12-28
Add update-site for scalariform

sschaef:
This change works in the same way as the change for the
scala-refactoring update-site in commit b6ed45040c1.


sschaef:
Build failure is expected. scalariform is still source incompatible.

dragos:
What should we do with this PR?

sschaef:
Probably best when it is closed. We first need to restore compatibility of the newest scalariform branch to the latest scalariform version that was published by mdr. Since I do not have the time to fix all of the problems in scalariform right now we shouldn't merge this. Users wouldn't like any incompatibilities.

PR #1055, opened 2015-12-30
Introduces barrier for Hot Code Replacement

wpopielarski:
Scenario:
Hot Code Replacement feature in its flow calls
`ScalaHotCodeReplacementManager.doHotCodeReplace` method.
This method sets `ScalaDebugTarget.isPerformingHotCodeReplace` flag
which controls dropping stack trace top in `ScalaThread` instance
(see its `dropToFrameInternal`). And if more then one thread is doing
this action it happens sometimes that `ScalaThread.canDropToFrame`
does not allow for a drop. The root cause probably is the situation
when earlier thread set `isPerformingHotCodeReplace` flag to false
and later thread is just before `dropToFrameInternal`. The barrier
prevents this race condition.

### DIFF
--- a/org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/hcr/ScalaHotCodeReplaceManager.scala
+++ b/org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/hcr/ScalaHotCodeReplaceManager.scala
@@ -3,6 +3,8 @@
  */
 package org.scalaide.debug.internal.hcr
 
+import java.util.concurrent.atomic.AtomicBoolean
+
 import scala.collection.JavaConverters
 import scala.collection.mutable.Publisher
 import scala.concurrent.ExecutionContext.Implicits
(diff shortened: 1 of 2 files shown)


org.scala-ide.sdt.debug/src/org/scalaide/debug/internal/hcr/ScalaHotCodeReplaceManager.scala
```diff
@@ -150,8 +152,17 @@ private[internal] trait HotCodeReplaceExecutor extends Publisher[HCRResult] with
     }
   }
 
+  protected val doHotCodeReplaceBarrier = new AtomicBoolean
+  private def waitForBarrier(): Unit = {
+    if (doHotCodeReplaceBarrier.getAndSet(true)) {
+      while (doHotCodeReplaceBarrier.get) {}
```
sschaef: What do you expect, how long would this loop take? Is the whole race condition happening in < 1ms or in something that is way larger?

wpopielarski: actually, it is very rare situation and I think related to test rather not real user work. I do not have any statistics.

sschaef:
LGTM

PR #1062, opened 2016-01-14
Fix NPE in EclipseLogger

sschaef:


sschaef:
@wpopielarski Can you see any other way how this NPE could occur?

sschaef:
retest this please

org.scala-ide.sdt.core/src/org/scalaide/core/internal/logging/EclipseLogger.scala
```diff
@@ -100,8 +100,10 @@ object EclipseLogger extends Logger {
 
   // Because of a potential deadlock in the Eclipse internals (look at #1000914), the log action need to be executed in the UI thread.
   private def logInUiThread(severity: Int, message: Any, exception: Throwable): Unit = {
-    val status = new Status(severity, IScalaPlugin().getBundle.getSymbolicName, message.toString, exception)
-    if (IScalaPlugin().headlessMode) pluginLogger.log(status)
-    else DisplayThread.asyncExec { pluginLogger.log(status) }
+    if (message != null || exception != null) {
```
wpopielarski: not sure but I think it makes sense to make message or throwable mandatory like:
```
private def logInUiThread(...): Unit = {
assert (message != null || exception != null)
...
```
I think that situation when message and throwable are null is incorrect, but maybe I'm wrong

Kwestor: I think that breaking user code in case he provided incorrect log message is not a good idea.

sschaef: I agree with Jerzy, we shouldn't throw exceptions here. A good idea as I think is that we could log an error that says that an error occurred while logging an error. ;)

wpopielarski: upps, you're right, I forgot the fact that it is helper function so should
not break the execution. Simon's solution is quite ok for me, just I'm
wonder how it is solved in other logging libs?

2016-01-14 19:01 GMT+01:00 Simon Schäfer <notifications@github.com>:

wpopielarski: ok, I made a glimpse of eye on log4j and looks like they don't worry about
null both message and throwable. So current code LGTM :)

2016-01-15 9:18 GMT+01:00 Wieslaw Popielarski <wpopielarski@gmail.com>:

sschaef: Alright, I added another commit, which logs an error in case message and exception are both null.

wpopielarski:
LGTM

PR #1064, opened 2016-01-21
Remove unused imports

sschaef:
These imports were not seen so far, since the compiler doesn't report
them in the most actual nightly build of scalac. However, the newest
version of scalac includes the following fix: https://github.com/scala/scala/pull/4899

If the IDE is build on top of a scalac SNAPSHOT that includes this fix,
then it reports the unused imports that are removed by this commit.


sschaef:
Merge without review because I need this in master.

PR #1066, opened 2016-01-26
Hcr flaky tests more data

wpopielarski:
Collects more data about result of HCR test execution


sschaef:
Please rebase onto master

sschaef:
This PR is of no use to find out where the problem comes from. I just let the tests disabled until they magically fix themselves.

PR #1067, opened 2016-01-29
Fix early initialization bug in AbstractMemberProposal

sschaef:
In the previous commit that changed that file I did a change which is
reverted by this commit. I thought that `new { ... } with X` is
equivalent to `new X { ... }`, which is not the case. The former uses
early initialization in order to set values in the trait `X`, whereas
the latter just overrides the values.

Fix #1002642


dragos:
LGTM (I didn't run the code though, but given your explanations I think this is the right thing to do).

wpopielarski:
looks like running this piece could cause NPE. on other hand I don't like
this kind of construction. This abstraction should get object as
constructor parameter like AbstractMemberProposal(abstractMethod: ...)

2016-01-31 18:15 GMT+01:00 Iulian Dragos <notifications@github.com>:

sschaef:
On 01/31/2016 08:10 PM, Wiesław Popielarski wrote:
> looks like running this piece could cause NPE. on other hand I don't like
> this kind of construction. This abstraction should get object as
> constructor parameter like AbstractMemberProposal(abstractMethod: ...)
I'm too lazy to add this kind of refactoring now. Also, this is only 
used inside of one file inside of the internal API, therefore we don't 
have to think about it again.

PR #1068, opened 2016-02-12
Retain async history should not break debugger

wpopielarski:
adding frames to async history should not break debugger. We just will
lose this record from memory.


wpopielarski:
1. messageOrdinal is updated only if `frames()` is succeeding so that's why it's lazy.
2. I wish I would wrap it in `Try` but unfortunately JDI exception is turned into ugly `InternalError`, so bad luck

sschaef:
LGTM

PR #1071, opened 2016-02-22
Store generated sources in Scala IDE source folder

sschaef:
The extension compiler will automatically load these files. The runtime
generated classes will no longer be considered. This makes the code against binary incompatibilities and the load times faster.

Fixes #1002652


sschaef:
@wpopielarski I would like to get your go

org.scala-ide.sdt.core/src/org/scalaide/core/internal/extensions/ExtensionCompiler.scala
```diff
@@ -359,3 +381,38 @@ object ExtensionCompiler extends AnyRef with HasLogger {
...
+/**
+ * Generator for the creator classes of the IDE extensions. Do not forget to
+ * specify the `saveLocation` before you run this App. It needs to point to
+ * something like `/path/to/scala-ide/org.scala-ide.sdt.core/src`. All the
+ * generated sources are written together with their package hierarchy into this
+ * directory.
+ */
+object ExtensionGenerator extends App {
+
+  private val saveLocation = "/home/antoras/dev/scala/scala-ide/org.scala-ide.sdt.core/src"
```
dragos: Can this be empty or passed as a command line argument?

dragos: Actually, how do you run it? :) It doesn't seem to be part of the build, but it depends on the rest of the IDE :)

dragos:
It looks fine, but I'd like to make sure someone else can generate those sources as well ;-). @wpopielarski it would be good if you took a look.

org.scala-ide.sdt.core/src/org/scalaide/core/internal/extensions/ExtensionCompiler.scala
sschaef: the path is an oversight, it should be removed from the sources. Beside from that, the generator extends App, therefore it can be run by running its main method. Since we only need to regenerate theses sources when we change the code that needs to be generated, I created it as an App. We could also introduce it into the build but I guess the way through the main method is the easiest.

dragos: Yeah, but as I said: it depends on `sdt.core` and the whole eclipse. How do you ran the `App`? What classpath did you pass?

sschaef: I ran it in Eclipse, which figured out the classpath by itself. Since this app is part of the sdt.core project it shares the same classpath.

wpopielarski: ok, I try it on my local too

2016-02-24 11:03 GMT+01:00 Simon Schäfer <notifications@github.com>:

wpopielarski:
just a question if it is run from /org.scala-ide.sdt.core project the `saveLocation` cannot be project dependent?

wpopielarski:
Funny, I just realize that this generator could be a part of something scala-ide-tools bundle and available in eclipse as maybe button only in case /org.scala-ide.sdt.core is in the set of open projects :D but I'm aware too much work or in Polish proverb a hide is not worth to be tanned.

sschaef:
`saveLocation` can be anything actually. Only the app needs to be run from within of Eclipse (or one needs to create the classpath manually).

sschaef:
I removed the hardcoded location string from the sources.

PR #1073, opened 2016-02-24
 Allow users to specify the spacing around block imports

sschaef:
A new preference option is added to the Scala > Editor preference page,
which allows users to configure if they want to see spaces around a
block import. The default configuration is that the spaces are added,
because this is the behavior of scalariform (whose behavior is hardcoded
and can't be changed).

Fix #1001692

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/ui/internal/preferences/EditorPreferencePage.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/ui/internal/preferences/EditorPreferencePage.scala
@@ -65,6 +65,9 @@ class EditorPreferencePage extends PreferencePage with IWorkbenchPreferencePage 
 
     val outline = group("Outline", base)
     checkBox(P_INITIAL_IMPORT_FOLD, "Fold import nodes by default", outline)
+
+    val refactoring = group("Refactoring", base)
+    checkBox(SPACES_AROUND_IMPORT_BLOCKS, "Add spaces around braces in multi-import block", refactoring)
 }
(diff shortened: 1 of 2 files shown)


org.scala-ide.sdt.core/src/org/scalaide/ui/internal/preferences/EditorPreferencePage.scala
```diff
@@ -101,24 +104,24 @@ class EditorPreferencePage extends PreferencePage with IWorkbenchPreferencePage
...
-  final val P_ENABLE_AUTO_BREAKING_COMMENTS = BASE + "autoBreakingComments"
-
-  final val P_ENABLE_MARK_OCCURRENCES = BASE + "markOccurences"
-  final val P_SHOW_INFERRED_SEMICOLONS = BASE + "showInferredSemicolons"
-
-  final val INDENT_GUIDE_ENABLE = BASE + "indentGuideEnable"
-  final val INDENT_GUIDE_COLOR = BASE + "indentGuideColor"
-  final val P_ENABLE_HOF_COMPLETION = BASE + "completionAlwaysLambdas"
-  final val P_INITIAL_IMPORT_FOLD = BASE + "initialImportFold"
+  final val P_ENABLE_AUTO_CLOSING_COMMENTS = "scala.tools.eclipse.editor.autoClosingComments"
```
dragos: Okay...

dragos:
@wpopielarski ?

wpopielarski:
LGTM

PR #1074, opened 2016-02-26
Add Mars profile for building

dragos:
This makes the IDE compile under Eclipse 4.5 (Mars).

- added `eclipse-mars` profile in pom.xml, with proper dependency on the weaving.hook
- source changes for making it compile with Mars
- duplicated sources from `src-luna` to `src-mars` in the `sdt.core` and `sdt.aspects` projects

- [ ] check if we need to keep different source folders, or we can drop Luna altogether (maybe the build is binary compatible?)

### DIFF
--- a/org.scala-ide.sdt.aspects/src-mars/scala/tools/eclipse/contribution/weaving/jdt/cfprovider/ClassFileProviderAspect.aj
+++ b/org.scala-ide.sdt.aspects/src-mars/scala/tools/eclipse/contribution/weaving/jdt/cfprovider/ClassFileProviderAspect.aj
@@ -1,0 +1,352 @@
+/*
+ * Copyright 2005-2010 LAMP/EPFL
+ */
+// $Id$
+
+package scala.tools.eclipse.contribution.weaving.jdt.cfprovider;
+
+import java.util.HashSet;
+
+import org.eclipse.core.filebuffers.FileBuffers;
+import org.eclipse.core.filebuffers.ITextFileBuffer;
+import org.eclipse.core.filebuffers.ITextFileBufferManager;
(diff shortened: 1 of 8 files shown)


dragos:
I won't be able to work more on this, but this builds fine locally (and tests pass). I hope we can move our dev to Mars, since it's better than Luna, and that's what I'm using day to day.

dragos:
retest this please

dragos:
OK, the tests are failing now because of scalariform not being found as a P2 repository. @sschaef how is it supposed to work now? Via Maven snapshots? Then should we just remove the P2 repo from `pom.xml`?

wpopielarski:
I'm not sure about that because I'm using luna by now, but Krzysztof alluded that mars uses same error icon for markers coming from compilation and presentation compiler. I need to verify it

dragos:
I'm using Mars for a while now.. and I didn't notice anything.. These changes are only for being able to build the source code under Mars (the binary already runs under Mars). This is of benefit only for IDE developers, if they want to use the most recent Eclipse ;-)

sschaef:
There are source incompatible changes between Luna and Mars. We can't just have one source folder, for both platforms.

dragos:
Eclipse Mars passed! \o/

screen shot 2016-02-27 at 15 50 28

dragos:
No, but maybe we don't have to. If the binaries run on Luna as well (to be checked, but in principle they should), we can just simplify the build and get rid of special source folders.

The good news is that the Maven build and tests pass, and the weaving hook version is the right one.

sschaef:
Ok, in this case I also have to start to use Mars.

dragos:
Closing this, please follow #1078

PR #1075, opened 2016-02-26
Temporarily add new resolvers for scalariform and equinox weaving

sschaef:
The URL download.scala-ide.org is temporarily not available.


sschaef:
The build succeeded. Error is irrelevant since it is due to a config change on Jenkins.

PR #1076, opened 2016-02-28
Fix memory leak

sschaef:
See ticket [#1002655](https://www.assembla.com/spaces/scala-ide/tickets/1002655-memory-leak-in-scalareconciler/details#)

I'm not 100% sure if this really fixes the memory leak that is reported by MAT but it fixes at least one memory leak. In general we added entries to a preference store on editor creation but didn't remove the entries once the editor is closed. I believe that this is the memory leak that is reported by MAT because the class `BaseSemanticAction` keeps a reference to the editor, which means that not only the `BaseSemanticAction` object itself is kept in memory but also the `ScalaSourceFileEditor`, which is the problematic one. The former is just a small object, whereas the latter keeps lots of compilation units alive and therefore eats memory like shit. See commit messages for a little bit more information about the concrete refactorings.


wpopielarski:
how I see it the `_listener` should be kept as long as the counterpart instance of `BaseSemanticAction` is referred by strong reference.
secondly, gc doesn't have to clean weak refs immediately, there is the example of one guy on stackoverflow where it came into the real after 20 gc call.

sschaef:
Even if this doesn't fix the problem which I encountered, at least we are no longer depending on this WeakReference stuff. It should only be used if there is no other way to solve the problem. The new code, which explicitly disposes the listeners is way easier to understand.

wpopielarski:
agree with you Simon, `dispose` should be smooth and clean. i'm just wonder what about with `implicitHighlighter` in `ScalaClassFileEditor`

sschaef:
Good point, forgot about that one. I added it.

wpopielarski:
LGTM

PR #1077, opened 2016-02-29
Increase timeout for Futures in test suite

sschaef:
The Scala testing machine seems to be slower than our own machine (or
they have more load than we have). It happened now multiple times that
tests failed because of a timeout. Increasing the timeout should fix the
problem.

See for example

  https://scala-ci.typesafe.com/job/scala-2.11.x-integrate-ide/2195/console


wpopielarski:
LGTM (i I were you I would write 1 second, but 1000 millis is same good)

PR #1080, opened 2016-03-01
Select first match in quick outline

dragos:
This has been annoying me for a while.. Quick outline is now:

- case insensitive
- a matching definition is automatically selected, so you can just press Enter and Go!


dragos:
The `eclipse-mars` profile will fail, that's normal, but the Luna one is fine

sschaef:
LGTM

PR #1083, opened 2016-03-09
Update Scala IDE version to 4.5.0

sschaef:


mkurz:
Will this still come with Luna by default?

sschaef:
On 03/09/2016 07:16 PM, Matthias Kurz wrote:
>
> Will this still come with Luna by default?
>
So far, yes. But there is still time to update to Mars if this is 
important for you.

mkurz:
@sschaef If it isn't a big hassle, then it would be nice.

PR #1085, opened 2016-03-14
Do not use maven.build.timestamp for OSGi version of sbt project

sschaef:
Normally the usage of `maven.build.timestamp` seems to result in a
correct OSGi version but as the ticket shows, it also can result in an
invalid version. The problem is fixed by hardcoding the format of the
generated timestamp to `yyyyMMddHHmm`. Furthermore we also append the
short git hash to the version. The git hash is not required but may make
it easier to find out against which sources the plugin is built.

Fix #1002667

Review by @dragos


sschaef:
The new version number looks like this:
```
Bundle-Version: 0.13.8.local-2_11-201603141650-ddf46d0
```

PR #1090, opened 2016-04-04
Save only a position instead of a compiler symbol in implicit annotat…

dragos:
…ions.

Compiler symbols should not be kept inside UI elements or other long-lived objects. Symbols may retain in memory a full Scala compiler instance, including all the other symbols and types, leading to memory leaks. In this commit I removed the reference to the implicit symbol, instead keeping only the position. In case the user wants to hyperlink, we re-compute the symbol under the given position, and produce a hyperlink.

Fix #1002679

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/decorators/implicits/ImplicitConversionsOrArgsAnnotation.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/decorators/implicits/ImplicitConversionsOrArgsAnnotation.scala
@@ -3,8 +3,15 @@
  */
 package org.scalaide.ui.internal.editor.decorators.implicits
 
+import scala.reflect.internal.util.Position
+
+import org.eclipse.jface.text.IRegion
 import org.eclipse.jface.text.hyperlink.IHyperlink
 import org.eclipse.jface.text.source.Annotation
+import org.scalaide.core.compiler.IScalaPresentationCompiler.Implicits.RichResponse
+import org.scalaide.core.compiler.InteractiveCompilationUnit
+import org.scalaide.logging.HasLogger
+import org.scalaide.ui.editor.InteractiveCompilationUnitEditor
 import org.scalaide.ui.editor.ScalaEditorAnnotation
 
 object ImplicitAnnotation {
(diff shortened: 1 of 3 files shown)


org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/decorators/implicits/ImplicitConversionsOrArgsAnnotation.scala
```diff
@@ -14,8 +21,42 @@ object ImplicitAnnotation {
...
+    icu.withSourceFile { (sourcefile, compiler) =>
+      import compiler.ApplyImplicitView
+
+      /* Return a range position for the given line without leading whitespace. */
+      def trimLinePos(line: Int) = {
+        var offset = sourcefile.lineToOffset(line)
+        while (sourcefile.content(offset).isWhitespace)
+          offset += 1
+        val start = offset
+        while (sourcefile.content(offset) != '\n')
```
wpopielarski: isn't safer to use `Properties.lineSeparator`?

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/decorators/implicits/ImplicitHighlightingPresenter.scala
```diff
@@ -55,13 +55,12 @@ object ImplicitHighlightingPresenter {
...
       val region = new Region(pos.offset, pos.getLength)
-      val msg = compiler.asyncExec{
+      val msg = compiler.asyncExec {
         val sname = t.fun.symbol.nameString
         s"Implicit conversion found: `$txt`$DisplayStringSeparator`$sname($txt): ${t.tpe}`"
       }.getOption()
 
-      val annotation = new ImplicitConversionAnnotation(
-          () => compiler.mkHyperlink(t.symbol, name = "Open Implicit", region, scu.scalaProject.javaProject), msg.getOrElse(""))
+      val annotation = new ImplicitConversionAnnotation(t.pos, region, msg.getOrElse(""))
```
wpopielarski: so the main problem comes from the fact that `symbol` is quite heavy (I mean other objects can refer it too)?

wpopielarski:
LGTM

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/decorators/implicits/ImplicitHighlightingPresenter.scala
dragos: Yes, the main problem is that the symbol refers to many other objects (it's nested, and has `outer` and `owner` fields), which really may keep a full compiler in memory. The ticket has some screenshots with memory heap dumps.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/decorators/implicits/ImplicitConversionsOrArgsAnnotation.scala
dragos: Yes, that's a better option. I'll see how we do it in auto-edits

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/editor/decorators/implicits/ImplicitConversionsOrArgsAnnotation.scala
```diff
@@ -35,7 +35,7 @@ class ImplicitConversionAnnotation(pos: Position, region: IRegion, text: String)
         while (sourcefile.content(offset).isWhitespace)
           offset += 1
         val start = offset
-        while (sourcefile.content(offset) != '\n')
+        while (!sourcefile.isEndOfLine(offset))
```
wpopielarski: +1

PR #1091, opened 2016-04-05
Override Indicators now don’t hold on to compiler symbols.

dragos:
This should remove one source of memory leaks. Also, only create one
override indicator per definition, towards the closest override.

Fix #1002293

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/ui/internal/jdt/model/ScalaOverrideIndicatorBuilder.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/ui/internal/jdt/model/ScalaOverrideIndicatorBuilder.scala
@@ -1,17 +1,21 @@
 package org.scalaide.ui.internal.jdt.model
 
 import java.util.{ Map => JMap }
+
+import scala.reflect.internal.util.Position
+import scala.tools.eclipse.contribution.weaving.jdt.IScalaOverrideIndicator
+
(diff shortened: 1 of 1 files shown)


org.scala-ide.sdt.core/src/org/scalaide/ui/internal/jdt/model/ScalaOverrideIndicatorBuilder.scala
```diff
@@ -35,28 +39,68 @@ case class JavaIndicator(scu: ScalaCompilationUnit,
   }
 }
 
-trait ScalaOverrideIndicatorBuilder { self : ScalaPresentationCompiler =>
```
wpopielarski: just a question, what was the reason to make `ScalaIndicator` a part of this trait? I see that presentation compiler can be retrieved from `ScalaCompilationUnit`.

wpopielarski:
LGTM

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/jdt/model/ScalaOverrideIndicatorBuilder.scala
dragos: Yes, I don't know either. I think the problem was the constructor parameter `base: Symbol`. That type is nested inside the compiler.

dragos:
Thanks @wpopielarski! merging this.

PR #1092, opened 2016-04-08
Fix src-luna source folders in classpath files

sschaef:
The correct source folders weren't added in the PR that introduced the
Mars build and therefore we neither do have a out of the box working
Luna or Mars setup for Eclipse.


wpopielarski:
LTGM

PR #1093, opened 2016-04-09
Depend on final Scala version in nightly builds

sschaef:
The Scala people don't publish nightlies of Scala anymore, therefore we
have to depend to the latest Scala release which is the most actual
published Scala version.


dragos:
I don't know about the scala-refactoring change, but the scala version number is passed form the outside `-Dscala211.version`, so the nightly build doesn't rely on this default (it may make life easier for developers though, since 2.11.8-SNAPSHOT is not available on Sonatype anymore)

sschaef:
I also changed the configuration in Jenkins.

PR #913, opened 2015-03-16
Simple Sbt editor

dragos:
# Do not merge

This is in preparation for my ScalaDays talk. It shows how to create a functional Sbt editor using 4.x APIs. It needs changes in #912 to look nice.

# Help needed

- [ ] fix hard-coded paths
- [ ] add jars coming from sbt plugins

PR #1095, opened 2016-04-11
Sbt integration

sschaef:
This merges https://github.com/scala-ide/sbt-integration into scala-ide core. Furthermore it also integrates @dragos first attempt of an sbt editor (in #912 and #913) into the new module `sbt.core`.

TODO:

- [ ] Replace hard-coded paths in `SbtPresentationCompiler` with sbt-server integration.
- [ ] Remove `DefaultScalaEditorConfiguration`. This class simply shouldn't exist. We should remove functionality that can't be used in the sbt editor out of `ScalaSourceViewerConfiguration` and not add another source viewer configuration that doesn't have any functionality.


org.scala-ide.sdt.core/src/org/scalaide/ui/editor/DefaultScalaEditorConfiguration.scala
```diff
@@ -0,0 +1,107 @@
...
+ *
+ *  The workhorse of this implementation is [[InteractiveCompilationUnit]], which provides
+ *  the mapping between original source and Scala-source.
+ *
+ *  Any method can be overwritten.
+ *
+ *  @note Standard content assist is missing from this configuration, but should be easy to add
+ *        in a subsequent release
+ *
+ *  @since 4.3.0
```
dragos: I think this should be 4.5.0

org.scala-ide.sbt.core/pom.xml
```diff
@@ -0,0 +1,182 @@
...
+      <version>${pickling.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.typesafe.akka</groupId>
+      <artifactId>akka-stream_2.11</artifactId>
+      <version>${akka.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.typesafe.akka</groupId>
+      <artifactId>akka-actor_2.11</artifactId>
```
dragos: Shouldn't this be `akka-actor_${scala.binary.version}`? We may need to build the IDE for 2.12 soon.

org.scala-ide.sbt.core/src/org/scalaide/sbt/ui/internal/editor/SbtCompilationUnit.scala
```diff
@@ -0,0 +1,133 @@
...
+  override val scalaSource =
+    s"""|$prefix
+        |${originalSource.mkString("")} }}""".stripMargin.toCharArray()
+
+  override lazy val sourceFile = new BatchSourceFile(file, scalaSource)
+
+  override val scalaPos = ScalaPosition
+  override val originalPos = OriginalPosition
+
+  override def scalaLine(line: Int): Int = line + 3
```
dragos: Maybe make this magic constant be the number of lines in `$prefix`, since any change in the prefix would break this code.

dragos:
It looks like Akka requires Java 8...

sschaef:
On 04/12/2016 02:20 PM, Iulian Dragos wrote:
>
> It looks like Akka requires Java 8...
>
Yes, we need to update the Java version anyway for Mars. Can we update 
the Java version on Jenkins?

dragos:
I'll look into it.

org.scala-ide.sdt.core/src/org/scalaide/ui/editor/DefaultScalaEditorConfiguration.scala
```diff
@@ -0,0 +1,107 @@
...
+ */
+trait DefaultScalaEditorConfiguration extends SourceViewerConfiguration {
```
dragos: BTW, I think the existing `ScalaSourceViewerConfiguration` should subclass this class. But I don't think we can assume all Scala-based editors can inherit *all* of the existing Scala configuration. The Scala editor inherits the Java one, and some things will be done differently between Scala and Sbt files (or Play templates). I don't see how they can all share the exact same class, but if you manage, that's be great. But it's way easier to just refactor common configuration into this one, and inherit it elsewhere.

sschaef: Yes, I agree. Extracting common logic into one base trait is my plan.

dragos:
retest this please

dragos:
Seems that Java 8 is in place. Only one test failed, on Mars. Probably a flaky test.

sschaef:
Ok, thanks.

sschaef:
Superseded by the scala-ide/sbt-integration branch

PR #1097, opened 2016-04-28
Replaces AtomicReference

wpopielarski:
with oldfashioned synchronized because for threads contition spinning causes CPU overheating

fix #1002690


sschaef:
retest this please

sschaef:
The methods are synchronizes but wouldn't it be possible that multiple threads access different methods and therefore there could be a race condition when accessing `threads`?

sschaef:
LGTM. The error is related to EE.

wpopielarski:
Simon,

generally access to 'threads' should be atomic (so synchronious in this
time). AtomicReference was working pretty good. I think that main problem
was related to iulian's finding so :+ used instead of :: and permanent
recreation of 'prev :+ threadToAdd'. 'synchronized' just does it once but
blocks bigger range of code.
this is how I see it. I hope Iulian can correct me if I'm wrong.

2016-04-28 15:07 GMT+02:00 Simon Schäfer <notifications@github.com>:

PR #1098, opened 2016-05-04
Tests @BeforeClass for flaky ImplicitTest

wpopielarski:
PR should overcome problems with flaky tests in Expression Evaluator


wpopielarski:
@Before/@After is not a problem. Closing it...

PR #1100, opened 2016-06-03
Sort type completions differently

mlangc:
This PR addresses the problems explained in detail in [#1002686](https://www.assembla.com/spaces/scala-ide/tickets/1002686-scala-type-completions-should-be-sorted-differently/details#). Note we might want to expose `ProposalRelevanceCfg` in the GUI so that users can tweak it individually in future versions. At the moment a hardcoded default configuration is used.

### DIFF
--- a/build-all.sh
+++ b/build-all.sh
@@ -9,8 +9,8 @@ ROOT_DIR=${PWD}
 
 if [ -z "$*" ]
 then
-  MVN_ARGS="-Pscala-2.11.x -Peclipse-luna clean install"
-  MVN_P2_ARGS="-Dtycho.localArtifacts=ignore -Pscala-2.11.x -Peclipse-luna clean verify"
+  MVN_ARGS="-Pscala-2.11.x -Peclipse-mars $ADDITIONAL_MVN_OPTS clean install"
+  MVN_P2_ARGS="-Dtycho.localArtifacts=ignore -Pscala-2.11.x -Peclipse-mars $ADDITIONAL_MVN_OPTS clean verify"
 else
   MVN_ARGS="$*"
   MVN_P2_ARGS="-Dtycho.localArtifacts=ignore $*"
--- a/org.scala-ide.sdt.core/src/org/scalaide/core/completion/ProposalRelevanceCalculator.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/core/completion/ProposalRelevanceCalculator.scala
@@ -49,7 +49,7 @@ class ProposalRelevanceCalculator(cfg: ProposalRelevanceCfg = DefaultProposalRel
   }
 
   def forJdtType(prefix: String, name: String): Int = {
-    val maxRelevance = MaxExternalRelevance / 4 * 3
+    val baseRelevance = MaxExternalRelevance / 4 * 3
 
     def deltaForPrefix(deltaIfMatch: Int, regexes: Seq[Regex]): Int = {
       regexes.foldLeft(0) { (acc, rx) =>
@@ -80,6 +80,6 @@ class ProposalRelevanceCalculator(cfg: ProposalRelevanceCfg = DefaultProposalRel
       name.length*3 +
       nestingLevel
 
-    internalToExternalRelevance(maxRelevance + bonus - penalty)
+    internalToExternalRelevance(baseRelevance + bonus - penalty)
   }
 }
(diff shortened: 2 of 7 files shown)


build-all.sh
```diff
@@ -9,8 +9,8 @@ ROOT_DIR=${PWD}
 
 if [ -z "$*" ]
 then
-  MVN_ARGS="-Pscala-2.11.x -Peclipse-luna clean install"
-  MVN_P2_ARGS="-Dtycho.localArtifacts=ignore -Pscala-2.11.x -Peclipse-luna clean verify"
+  MVN_ARGS="-Pscala-2.11.x -Peclipse-mars $ADDITIONAL_MVN_OPTS clean install"
+  MVN_P2_ARGS="-Dtycho.localArtifacts=ignore -Pscala-2.11.x -Peclipse-mars $ADDITIONAL_MVN_OPTS clean verify"
```
wpopielarski: haha I was wonder who first loses his patience to add it. The previous solution impelled to write all options from scratch I guess

org.scala-ide.sdt.core/src/org/scalaide/core/completion/ProposalRelevanceCalculator.scala
```diff
@@ -0,0 +1,85 @@
...
+    else (relevance * IntToExtRatio).round.toInt
+  }
+}
+
+class ProposalRelevanceCalculator(cfg: ProposalRelevanceCfg = DefaultProposalRelevanceCfg) extends HasLogger {
+  import ProposalRelevanceCalculator._
+
+  def forScala[CompilerT <: Global](pc: CompilerT)(prefix: String, name: String, sym: pc.Symbol, viaView: pc.Symbol, inherited: Option[Boolean]): Int = {
+    // rudimentary relevance, place own members before inherited ones, and before view-provided ones
+    var relevance = MaxInternalRelevance
```
wpopielarski: general remark only. what are these numbers? I get that they are arbitrary just to keep a space between different cases. I'm just thinking about something like:
```scala
if (...) relevance.downgradeToGlobal
if (...) relevance.downgradeToAccessor
```
but this is NOT the must, just suggestion

org.scala-ide.sdt.core/src/org/scalaide/core/completion/ProposalRelevanceCalculator.scala
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    val casePenalty = if (name.substring(0, prefix.length) != prefix) 50 else 0
+    relevance -= casePenalty
+
+    internalToExternalRelevance(relevance)
+  }
+
+  def forJdtType(prefix: String, name: String): Int = {
+    val maxRelevance = MaxExternalRelevance / 4 * 3
```
wpopielarski: just to be sure. is it `MaxExternalRelevance * 0.75`?

org.scala-ide.sdt.core/src/org/scalaide/core/internal/compiler/ScalaPresentationCompiler.scala
```diff
@@ -398,30 +399,9 @@ class ScalaPresentationCompiler(private[compiler] val name: String, _settings: S
       } else name
     val container = sym.owner.enclClass.fullName
```
wpopielarski: oh, I'm sorry I see it is old code. Sorry @mlangc

wpopielarski:
Cool stuff, LGTM

org.scala-ide.sdt.core/src/org/scalaide/core/completion/ProposalRelevanceCalculator.scala
mlangc: You are right, the code is not from me, but giving all this magic numbers proper names would make sense indeed. I can take care of this in the days to come...

org.scala-ide.sdt.core/src/org/scalaide/core/completion/ProposalRelevanceCalculator.scala
mlangc: Yes, with the difference that `MaxExternalRelevance * 0.75` would be a double, while `MaxExternalRelevance / 4 * 3` has type `Int` (and might contain a truncation error that is not relevant here). The name `maxRelevance` is actually bad, and I've just changed it... also tell me if you would prefer `(MaxExternalRelevance * 0.75).toInt` which might be clearer for some readers.

build-all.sh
sschaef: Instead of the env varible you could just pass `"$@"` at the end, that would forward all parameters from `build-all.sh` to maven.

sschaef:
LGTM

PR #1102, opened 2016-06-24
Make Neon default build and remove Luna and Mars builds

sschaef:
Neon has been released and it is really good. I no longer see a reason why we should keep the Luna and Mars builds around. The Luna build was probably only used by me and the Mars build can be replaced with the Neon build.


sschaef:
I'm merging without review. @wpopielarski You are in vacation, so I can't wait for you to return. If it breaks your setup, you can complain once you are back. ;)

mkurz:
@sschaef Is the `master` branch using `Eclipse Neon.1a (4.6.1)` already?

sschaef:
It is compatible if that is what you meant. Technically, both 4.6 and 4.6.1 are consumed from the same update site, therefore there is nothing special what we have to do in order to make Scala IDE compatible to the latest minor version of Eclipse.

PR #1106, opened 2016-07-03
Do not cache the `MarkOccurrenceIndex`

mlangc:
Caching the `MarkOccurrenceIndex` is problematic, since this quickly leads
to an index, that is referencing an AST that is not identical to the most
recent one from the presentation compiler. Since equality for symbols is
implemented as reference equality, this is a problem even if the trees are
semantically equivalent, resulting in erratic behaviour as described in
ticket [#1002701](https://app.assembla.com/spaces/scala-ide/tickets/1002701-mark-occurrences-behaves-erratically).

I've experimented with a solution in the refactoring library, that performs
semantic equality comparisons on symbols, to work around this problem. This
worked as far as the trees where indeed semantically equivalent. In some cases
however, there where semantic differences in the trees, even though the source
code did not change, like positions being available in one one tree,
but absent in the other. Maybe this could be sorted out by digging deeper, but
in order to keep it simple, I'd rather not go this route unless not caching
actually results in noticeable performance problems.

Fix #1002701

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/core/internal/decorators/markoccurrences/ScalaOccurrencesFinder.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/core/internal/decorators/markoccurrences/ScalaOccurrencesFinder.scala
@@ -8,8 +8,6 @@ import scala.tools.refactoring.implementations.MarkOccurrences
 import org.eclipse.jface.text.IRegion
 import org.eclipse.jface.text.Region
 import org.scalaide.util.Utils
-import scala.ref.WeakReference
-import org.scalaide.core.compiler.IScalaPresentationCompiler
 import org.scalaide.core.compiler.IScalaPresentationCompiler.Implicits._
 
 case class Occurrences(name: String, locations: List[IRegion])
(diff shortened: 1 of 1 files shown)


sschaef:
LGTM

org.scala-ide.sdt.core/src/org/scalaide/core/internal/decorators/markoccurrences/ScalaOccurrencesFinder.scala
```diff
@@ -24,28 +22,6 @@ case class Occurrences(name: String, locations: List[IRegion])
...
   import ScalaOccurrencesFinder._
 
-  /* IMPORTANT:
-   * In the current implementation, all access to `indexCache` are confined to the Presentation
-   * Compiler thread (and this is why `indexCache` requires no synchronization policy).
-   */
-  private var indexCache: Option[TimestampedIndex] = None
-
-  private def getCachedIndex(lastModified: Long, currentCompiler: IScalaPresentationCompiler): Option[MarkOccurrencesIndex] = indexCache match {
-    case Some(TimestampedIndex(`lastModified`, WeakReference(index))) if index.global eq currentCompiler => Some(index)
```
dragos: How come this test is not enough to ensure consistency? Are you sure this actually fixes the problem you reported?

Also, keep an eye on performance. This cache was added for a reason. May not be a huge problem since the feature can be easily turned off, but then the default should be off.

dragos:
FYI: https://groups.google.com/d/msg/scala-ide-user/fmTwrowuvHc/kSmuBdQwopQJ

org.scala-ide.sdt.core/src/org/scalaide/core/internal/decorators/markoccurrences/ScalaOccurrencesFinder.scala
mlangc: Well, if we wanted a proper cache, we should at least have used a `SoftReference` here, since `WeakReferences` are completely ignored by the GC and cleared rather quickly. We had cache hit rates well below 80%, at least on my system. What happened for me was that the feature worked exactly when the GC cleared the reference, so that everything was recreated from scratch, but failed otherwise.

As described in the commit message, the test is not enough to ensure consistency... the source is compiled multiple times, into semantically mostly, but not entirely equal trees, resulting in all kinds of headaches. I don't think it's worth the effort to invest any more time in this, unless someone complains, or has a really good idea.

mlangc:
I didn't study this thread carefully, so correct me if I'm wrong, but it references performance problems in the refactoring library with named arguments, that might have affected `MarkOccurences`. This should not be an issue with the current version of the refactoring library (see https://github.com/scala-ide/scala-refactoring/commit/0d775c6e75e10aed81b685079ebf1ef9c5c11a5b).

org.scala-ide.sdt.core/src/org/scalaide/core/internal/decorators/markoccurrences/ScalaOccurrencesFinder.scala
dragos: Hm, I don't really understand how the test is not enough to make sure the AST is up to date. If the cache is cleared early it may lead to some additional recompilations, but that shouldn't lead to anything being incorrect (as you noticed). If it's cache is hit, it means the source hasn't changed, so why wouldn't the AST be correct? That's the part I don't get.

Regarding performance issues, I prefer to be on the careful side when something is computed on every cursor move, click, or key pressed. I just pointed out how this was a serious problem years ago, leading to this cache, that you now removed. It's fine, but be aware and know where to look if things turn sour ;-) The ultimate test used to be `Typers.scala`, which at more than 5000 LOC, pushes the IDE to its limits. :)

PR #1112, opened 2016-08-26
Do now pass compiler reference to reporter

sschaef:
This led to a NPE because the compiler reference can't be initialized
early enough. The reporter needs to be fully initialized in the ctor of
Global but Global is the compiler and therefore it can't be passed to
the reporter in its subclass.

Fix #1002733


wpopielarski:
LGTM

sschaef:
retest this please

PR #1113, opened 2016-08-27
Automatically restart PC in order to avoid false positive errors

sschaef:
Because the PC ofter shows false positives the idea is to automatically
restart the PC, which would get rid of the false positives. After a
restart all of the invalid state in the PC is gone. A better fix would
be to not get invalid state in the PC but I don't understand the PC good
enough to provide such a fix.

The PC is automatically restart in the following situations:

- After Eclipse gained focus. This happens once Eclipse lost focus
  and gained it afterwards.
- After an editor gained focus. This happens when a different editor
  is selected.

Since restarting the PC is relatively fast and happens asynchronously in
the background, this shouldn't block users in any way.

Fix #1002725

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/core/internal/jdt/model/ScalaSourceFile.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/core/internal/jdt/model/ScalaSourceFile.scala
@@ -1,30 +1,33 @@
 package org.scalaide.core.internal.jdt.model
 
-import java.util.{ HashMap => JHashMap }
+import java.util.{ HashMap ⇒ JHashMap }
+
+import scala.tools.eclipse.contribution.weaving.jdt.IScalaSourceFile
+import scala.tools.nsc.interactive.Response
+import scala.tools.nsc.io.AbstractFile
+import scala.tools.nsc.io.VirtualFile
+import scala.util.control.Exception
+
 import org.eclipse.core.resources.IFile
+import org.eclipse.core.runtime.CoreException
+import org.eclipse.core.runtime.IPath
 import org.eclipse.core.runtime.IProgressMonitor
 import org.eclipse.jdt.core.ICompilationUnit
 import org.eclipse.jdt.core.IJavaElement
 import org.eclipse.jdt.core.IType
+import org.eclipse.jdt.core.JavaModelException
 import org.eclipse.jdt.core.WorkingCopyOwner
+import org.eclipse.jdt.core.compiler.CharOperation
 import org.eclipse.jdt.core.compiler.IProblem
-import org.eclipse.jdt.internal.core.util.HandleFactory
-import org.eclipse.jdt.internal.core.{ CompilationUnit => JDTCompilationUnit }
+import org.eclipse.jdt.core.dom.CompilationUnit
+import org.eclipse.jdt.internal.core.{ CompilationUnit ⇒ JDTCompilationUnit }
 import org.eclipse.jdt.internal.core.OpenableElementInfo
 import org.eclipse.jdt.internal.core.PackageFragment
-import scala.tools.nsc.io.AbstractFile
-import scala.tools.nsc.io.VirtualFile
-import scala.tools.eclipse.contribution.weaving.jdt.IScalaSourceFile
-import org.scalaide.core.resources.EclipseFile
-import org.eclipse.jdt.core.compiler.CharOperation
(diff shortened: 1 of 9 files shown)


wpopielarski:
LGTM

org.scala-ide.sdt.core/src/org/scalaide/core/internal/jdt/model/ScalaSourceFile.scala
```diff
@@ -33,58 +36,60 @@ class ScalaSourceFileProvider extends SourceFileProvider {
...
       None
     else {
       // Always `null` because `handleFactory.createOpenable` is only called to open source files, and the `scope` is not needed for this.
       val unusedScope = null
       val source = handleFactory.get().createOpenable(path, unusedScope)
       source match {
-        case ssf : ScalaSourceFile => Some(ssf)
-        case _ => None
+        case ssf: ScalaSourceFile ⇒ Some(ssf)
+        case _                    ⇒ None
```
wpopielarski: this formatting done by tool or you? generally I don't think this kind of formatting is cool

sschaef: auto format. I don't have anything against => adjusted to the same column if it is only one line, that makes it easier to read imho (especially if there are more than just two cases).

wpopielarski: LGTM

2016-08-30 11:47 GMT+02:00 Simon Schäfer <notifications@github.com>:

org.scala-ide.sdt.core/src/org/scalaide/core/internal/project/ScalaProject.scala
```diff
@@ -507,11 +518,6 @@ class ScalaProject private(val underlying: IProject) extends ClasspathManagement
     settings.javaextdirs.value = " "
     settings.classpath.value = (scalaCp.userCp ++ scalaCp.scalaLibrary.toSeq).map(_.toOSString).mkString(pathSeparator)
     scalaCp.scalaLibrary.foreach(scalaLib => settings.bootclasspath.value = scalaLib.toOSString)
-
-    logger.debug("javabootclasspath: " + settings.javabootclasspath.value)
```
dragos: Are these logged somewhere else now? They are pretty useful debugging info when the Java version isn't what you'd expect for a given project.

sschaef: Yes, after the PC is created, its settings are logged in PCProxy

org.scala-ide.sdt.core/src/org/scalaide/core/internal/project/ScalaProject.scala
```diff
@@ -67,15 +71,27 @@ object ScalaProject {
 
   /** Listen for [[IWorkbenchPart]] event and takes care of loading/discarding scala compilation units.*/
   private class ProjectPartListener(project: ScalaProject) extends PartAdapter with HasLogger {
+    override def partActivated(part: IWorkbenchPart): Unit = {
```
dragos: Does this restart the PC on every focus to the editor? So going through different editors will restart the PC each time, if there are errors?

dragos:
I'm pretty worried about this change. The PC isn't that fast to start, and a cold start means re-reading the JDK, the Scala library jar, etc. This adds significant memory churn, and switching between editors happens quite often. For example, a PC action like hyperlinking will be blocking until the PC is ready to take "orders". And if the error is genuine this will keep restarting the PC every time...

org.scala-ide.sdt.core/src/org/scalaide/core/internal/project/ScalaProject.scala
sschaef: yes, the PC will be restarted on every focus change if there are errors in the editor that is focused. I know this can hardly be described as a clean solution but I couldn't think of a better one. At least, the behavior is optional and can be disabled in the preferences.

sschaef:
Yes, we discussed that during our meeting. We will do thorough testing until we release the next final version. If it turns out that this behavior is indeed not useful in many situations, we will remove it.

dragos:
I installed the latest nightly, and even though I see the PC restarting often, it's not that bad so far. I'll keep an eye on it as well.

PR #1116, opened 2016-08-30
 Remove annotations that belong to a removed range of an editor

sschaef:
When text is removed from an editor the annotations that belong to the
removed range need to be removed too because they are no longer valid.

This is especially useful functionality when the user fixes an error. In
such a case, the builder did already create an annotation previously
that needs to be removed. With this change, it can happen that errors
are removed which were not yet fixed. In most cases, the PC will find
them once again and show them to users. In rare cases, the builder needs
to be run to find all problems. This is not optimal behavior but
probably the best what we can get in Scala because we can't run the
builder on every source change, it is simly to slow.

Fix #1002726

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/ui/editor/DecoratedInteractiveEditor.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/ui/editor/DecoratedInteractiveEditor.scala
@@ -9,14 +9,14 @@ import org.eclipse.jdt.internal.ui.javaeditor.CompilationUnitDocumentProvider.Pr
 import org.eclipse.jface.text.ITextViewerExtension2
 import org.eclipse.jface.text.Position
 import org.eclipse.jface.text.source.Annotation
-import org.eclipse.jface.text.source.IAnnotationModel
+import org.eclipse.jface.text.source.IAnnotationModelExtension2
 import org.scalaide.util.internal.eclipse.AnnotationUtils._
 import org.scalaide.util.ui.DisplayThread
 
 trait DecoratedInteractiveEditor extends ISourceViewerEditor {
 
   /** Return the annotation model associated with the current document. */
-  private def annotationModel = Option(getDocumentProvider).map(_.getAnnotationModel(getEditorInput).asInstanceOf[IAnnotationModel])
+  private def annotationModel = Option(getDocumentProvider).map(_.getAnnotationModel(getEditorInput))
 
   private var previousAnnotations = List[Annotation]()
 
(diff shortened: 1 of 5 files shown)


org.scala-ide.sdt.core/src/org/scalaide/ui/editor/DecoratedInteractiveEditor.scala
```diff
@@ -19,10 +21,21 @@ trait DecoratedInteractiveEditor extends ISourceViewerEditor {
   private var previousAnnotations = List[Annotation]()
 
   /**
+   * This removes all annotations in the region between `start` and `end`.
+   */
+  def removeAnnotationsInRegion(start: Int, end: Int): Unit = annotationModel foreach { model ⇒
+    val annsToRemove = model.getAnnotationIterator.asScala.filter { ann ⇒
```
dragos: You could use `IAnnotationModelExtension2.getAnnotationIterator(offset,length,...)`. It might be more efficient than linear search.

org.scala-ide.sdt.core/src/org/scalaide/ui/editor/DecoratedInteractiveEditor.scala
```diff
@@ -35,15 +48,19 @@ trait DecoratedInteractiveEditor extends ISourceViewerEditor {
...
     if (presViewer.isInstanceOf[ITextViewerExtension2]) {
-        // TODO: This should be replaced by a better modularization of semantic highlighting PositionsChange
-        val newPositions = newAnnotations.values
-        def end (x:Position) = x.offset + x.length - 1
-        val taintedBounds : (Int, Int) = ((Int.MaxValue, 0) /: newPositions) {(acc, p1) => (Math.min(acc._1, p1.offset), Math.max(acc._2, end(p1)))}
-        val taintedLength = (taintedBounds._2 - taintedBounds._1 +1)
+      // TODO: This should be replaced by a better modularization of semantic highlighting PositionsChange
+      val newPositions = newAnnotations.values
+      def end(x: Position) = x.offset + x.length - 1
+      val taintedBounds = (newPositions foldLeft (Int.MaxValue, 0)) { (acc, p1) ⇒ (Math.min(acc._1, p1.offset), Math.max(acc._2, end(p1))) }
```
wpopielarski: `(newPositions reduce { (z, e) => (Math.min(z._1, e.offset), Math.max(z._2, end(e...))` ?

wpopielarski: I have an impression that this computation is not needed. Isn't it just:
```scala
val startsAt = newPosition.minBy(_.offset).offset
val length = newPosition.maxBy(_.length).length
```
?

wpopielarski:
generally LGTM

org.scala-ide.sdt.core/src/org/scalaide/ui/editor/DecoratedInteractiveEditor.scala
```diff
@@ -35,15 +48,19 @@ trait DecoratedInteractiveEditor extends ISourceViewerEditor {
     // http://stackoverflow.com/questions/12507620/race-conditions-in-annotationmodel-error-annotations-lost-in-reconciler
     val presViewer = getViewer
```
wpopielarski: do we need this line? `presViewer` is actually not used. This is not the must of course.

org.scala-ide.sdt.core/src/org/scalaide/ui/editor/DecoratedInteractiveEditor.scala
sschaef: It is over my head right now.

org.scala-ide.sdt.core/src/org/scalaide/ui/internal/reconciliation/ScalaReconcilingStrategy.scala
```diff
@@ -35,7 +35,16 @@ class ScalaReconcilingStrategy(icuEditor: InteractiveCompilationUnitEditor) exte
   override def setProgressMonitor(pMonitor: IProgressMonitor): Unit = {}
 
   override def reconcile(dirtyRegion: DirtyRegion, subRegion: IRegion): Unit = {
```
sschaef: I just found out that once incremental reconciliation is enabled, the other method is not called anymore. I need to move its implementation.

sschaef:
Alright, I think all comments are addressed and the new incremental reconciliation now works as it should.

PR #1118, opened 2016-09-05
Revert introduction of new feature that can automatically restart PC

sschaef:
The automatic restarts of the PC were introduced in order to get rid of
false positive errors that are shown in the editor. After testing this
feature for some time in usage it turned out that this feature is
basically worthless.

The main problem is that most of the time an editor shown real errors or
warnings and in all of these cases the PC is restarted, even though it
shouldn't. There is no way to test if an error is a false positive or
not, therefore the PC is restarted many times, which not only slows down
the editor but also has weird effects. Semantic highlighting often is
cleared after the PC is restarted and therefore users have to live only
with syntax highlighting for a second or two. The automatic restarts
of the PC basically are an unsafe feature and should only be used
manually by the user. Syncing all of the possible semantic features of
the IDE with the restart times of the PC may be possible but seem to
require heavy efforts, which should not be invested in a feature that
after all is a hack to work around bugs in the PC.

Because of this, this new feature is removed.

Fix #1002741
Re #1002725


wpopielarski:
LGTM

PR #1119, opened 2016-09-12
Scalac java error ignore

wpopielarski:
Allows scalac ignores java errors. 

Java errors are reported to `SbtBuildReporter` by java compiler
directly.
Commit removes a bug when java depends on scala code (see `SbtInputs`)

Fixes #1002738


sschaef:
You are on the wrong branch. Right now you are on top of `sbt-integration` but you should be on top of `master`. `sbt-integration` is still not merged yet, therefore all commits need to be on the `master` branch (the exception of course is that you want to work on anything that was introduced in `sbt-integration`, in this case you need to open the PR against another branch. In the Github UI when you want to create a PR you can select the target branch).

wpopielarski:
superseded by #1120

wpopielarski:
superseded by #1120

PR #1120, opened 2016-09-13
Allows scalac ignores java errors.

wpopielarski:
Java errors are reported to `SbtBuildReporter` by java compiler
directly.
Commit removes a bug when java depends on scala code (see `SbtInputs`)

Fixes #1002738

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/core/internal/builder/zinc/SbtBuildReporter.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/core/internal/builder/zinc/SbtBuildReporter.scala
@@ -13,7 +13,7 @@ import org.scalaide.util.eclipse.FileUtils
 import org.scalaide.util.internal.SbtUtils
 
 private case class SbtProblem(severity: xsbti.Severity, message: String, position: xsbti.Position, category: String)
-  extends xsbti.Problem {
+    extends xsbti.Problem {
 
   import SbtUtils.m2o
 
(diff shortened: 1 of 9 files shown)


org.scala-ide.sdt.core/src/org/scalaide/core/internal/builder/zinc/SbtBuildReporter.scala
```diff
@@ -88,7 +109,7 @@ private[zinc] class SbtBuildReporter(project: IScalaProject) extends xsbti.Repor
       val markerPos = MarkerFactory.RegionPosition(offset, identifierLength(pos.lineContent, pos.pointer), line)
       BuildProblemMarker.create(resource, severity, msg, markerPos)
     } else
-      logger.error(s"suppressed error in Java file ${resource.getFullPath}:$line: $msg")
+      logger.warn(s"suppressed error in Java file ${resource.getFullPath}:$line: $msg")
```
sschaef: shouldn't it just log to `info` now? Afterwards javac reports the errors now as far as I understand.

org.scala-ide.sdt.core/src/org/scalaide/core/internal/builder/zinc/SbtInputs.scala
```diff
@@ -47,13 +47,15 @@ class SbtInputs(installation: IScalaInstallation,
...
+      Maybe.nothing[Analysis]
     else {
       val analysis = allProjects.collectFirst {
         case project if project.buildManager.buildManagerOf(f).nonEmpty =>
           project.buildManager.buildManagerOf(f).get.latestAnalysis(incOptions)
       }
-      Maybe.just(analysis.getOrElse(Analysis.Empty))
+      analysis.map { analysis =>
+        Maybe.just(analysis)
+      }.orElse(Option(Maybe.nothing[Analysis])).get
```
sschaef: `getOrElse` instead of `orElse.get`?

wpopielarski: agree but it was intentional. I thought rather about operating on same type. Doesn't matter, simplify it

sschaef:
LGTM

PR #1121, opened 2016-09-19
Ensure that mark occurrences is not called if it is disabled

sschaef:
There is no documentation on the `isMarkingOccurrences` method about
when it should be called. It was not called by the Java editor,
therefore we do it manually to ensure that mark occurrences is not
called when it is disabled.

Furthermore, the `isMarkingOccurrences` method is moved to
`ScalaCompilationUnitEditor` otherwise it can't be called in
`MarkOccurrencesEditorExtension`.


dragos:
Thanks a lot @sschaef. It was annoying every time it happened, but not annoying enough to look into a fix...

PR #1122, opened 2016-10-10
After migration to zinc all project is rebuilt

wpopielarski:
although the single file has been updated. The problem is touching
3 areas:
- class name finding in class jars on classpath, names of class files
  don't match with full class names, so class file names are transformed
  to class names
- zinc needs all classpath, so jdk libs are now added for analysis
- zinc tries to modify companions api entry in cache zip structure, so the
  lazy structures in analysis need to be materialized


sschaef:
Does this PR already fix [#1002746](https://app.assembla.com/spaces/scala-ide/tickets/1002746/details#)?

wpopielarski:
confirm

sschaef:
LGTM

PR #1124, opened 2016-10-19
Be insensitive for missing files

wpopielarski:
In multi projects compilation some files can be removed during
renaming as a result of recompilation request.

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/core/internal/builder/zinc/ProductExposer.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/core/internal/builder/zinc/ProductExposer.scala
@@ -8,14 +8,17 @@ import java.nio.file.Path
 import java.nio.file.SimpleFileVisitor
 import java.nio.file.StandardCopyOption
 
+import scala.util.Try
+
 import org.eclipse.core.resources.IProject
 import org.eclipse.core.resources.IResource
 import org.eclipse.core.runtime.IPath
 import org.eclipse.jdt.core.IJavaModelMarker
 import org.scalaide.core.IScalaPlugin
 import org.scalaide.core.SdtConstants
+import org.scalaide.logging.HasLogger
 
-private[zinc] object ProductExposer {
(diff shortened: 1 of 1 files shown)


org.scala-ide.sdt.core/src/org/scalaide/core/internal/builder/zinc/ProductExposer.scala
```diff
@@ -54,7 +55,12 @@ private[zinc] object ProductExposer {
               packagePath.listFiles(product(productPrefix))
           }.flatten.foreach { p =>
             val path = p.toPath
-            Files.move(path, path.resolveSibling(p.getName + ProductExtension), StandardCopyOption.REPLACE_EXISTING)
+            try {
+              Files.move(path, path.resolveSibling(p.getName + ProductExtension), StandardCopyOption.REPLACE_EXISTING)
+            } catch {
+              case e: Throwable =>
```
sschaef: You should only catch `NonFatal`.

org.scala-ide.sdt.core/src/org/scalaide/core/internal/builder/zinc/ProductExposer.scala
```diff
@@ -73,7 +79,12 @@ private[zinc] object ProductExposer {
       finder.collected
     }.foreach { path =>
       val fname = path.toFile.getName
-      Files.move(path, path.resolveSibling(fname.substring(0, fname.lastIndexOf(ProductExtension))), StandardCopyOption.REPLACE_EXISTING)
+      try {
+        Files.move(path, path.resolveSibling(fname.substring(0, fname.lastIndexOf(ProductExtension))), StandardCopyOption.REPLACE_EXISTING)
+      } catch {
+        case e: Throwable =>
+          logger.warn(s"IO error on show: ${e.getMessage} with resource ${path}")
```
sschaef: I think that you should mention in the log output that the exception is not fatal and that the operation simply could be skipped.

sschaef:
LGTM

PR #1127, opened 2016-11-06
Special handling type definitions in `ScalaStructureBuilder`

mlangc:
This tweak fixes [#1002773](https://app.assembla.com/spaces/scala-ide/tickets/1002773-open-type-fails-for-type-aliases/details#). I'm not sure that I like it though, so your comments are highly welcome!

### DIFF
--- a/org.scala-ide.sdt.core.tests/src/org/scalaide/core/findreferences/FindReferencesTests.scala
+++ b/org.scala-ide.sdt.core.tests/src/org/scalaide/core/findreferences/FindReferencesTests.scala
@@ -112,14 +112,15 @@ class FindReferencesTests extends FindReferencesTester with HasLogger {
 
   private def jdtElement2testElement(e: JavaElement): Element = {
     val testElement: String => Element = e match {
-      case e: ScalaDefElement      => Method.apply _
-      case e: ScalaAccessorElement => Method.apply _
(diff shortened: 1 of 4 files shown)


mlangc:
Hmm, I can't reproduce this test failure locally... retest this please!

sschaef:
retest this please

sschaef:
I have never worked with this part of the codebase. If it works, it is fine by me.

sschaef:
@mlangc What don't you like about this fix? maybe @wpopielarski wants to say something about it.

mlangc:
@sschaef Since I don't have any experience with the related code and with JDT stuff in general, I thought that there might be a more elegant way to fix this bug than adding two nodes of different types for the same element.

dragos:
LGTM, I think it's the minimal change that can support this use case.

wpopielarski:
Well, I would need to look into it deeper. No experience in this area. @mlangc not sure maybe some test would be helpful? If you don't have time I try to figure out something

wpopielarski:
generally area is interesting and I have a feeling that touches debugging model too (in phase when breakpoints are added)

wpopielarski:
give me some day to look into it :)

wpopielarski:
LGTM, @mlangc thanks for a hint how to fix "breakpoint not set" problem

PR #1129, opened 2016-11-16
Update zinc to 1.0.0-X6

sschaef:
There was a source incompatible change in SbtInputs and the classpath
file had to be fixed.


wpopielarski:
LGTM if build passes

PR #1131, opened 2016-11-30
Fixes problem of java/scala compilation

wpopielarski:
Right now project compilation is stopped when java compiler reports any
compilation problem. This is incorrect because actually problems have
different severity what had been overseen. So from now the compilation
is stopped when java compiler reports ERRORS and continued for all other
sorts of problems.


sschaef:
LGTM

PR #1136, opened 2017-01-05
Fix breakpoint in code

wpopielarski:
Currently there is no way to set a breakpoint in source code which file location does not correspond its package structure. This PR is going to fix this use case. The idea is to use Zinc analysis files to figure out the link between source and product. So it works for incremental build only.


sschaef:
LGTM. Are there any tickets that can be closed because of this?

PR #1146, opened 2017-03-10
Update to zinc-1.0.0 x9

mlangc:
Update zinc to 1.0.0-X9 (switching to X10 would require more refactorings).

Important: Go to `${WORKSPACE}/.metadata/.plugins/org.scala-ide.sdt.core` and
remove both `compiler-bridges/` and `compiler-interfaces/`. Otherwise the old
JARs stored in these directories might take precedence, leading to
`ClassNotFoundExceptions` during compiles.

This PR also contains a minor tweak to `ScalaPresentationCompiler` that is meant to help with debugging.


sschaef:
I have already updated to X10 in the 2.12 branch

wpopielarski:
just a remark: scala ide 2.12 is on X10 right now

mlangc:
Darn communication :-/ this cost me several hours... I should have asked.

sschaef:
The second commit is no longer needed, but the other two commits look legit. Even though I can't tell why the `nameLock` is no longer used.

PR #1153, opened 2017-04-06
Adds existential type to java type descriptor.

wpopielarski:


org.scala-ide.sdt.core/src/org/scalaide/core/internal/jdt/model/ScalaJavaMapper.scala
```diff
@@ -228,6 +229,7 @@ trait ScalaJavaMapper extends InternalCompilerServices with ScalaAnnotationHelpe
     val (sym, args) = tpe match {
       case TypeRef(_, sym, args) => (sym, args)
       case rt @ RefinedType(_, _) => (rt.typeSymbol, rt.typeArgs)
+      case ExistentialType(_, typ) => (typ.typeSymbol, Nil)
```
sschaef: shouldn't there be a default case `case _ =>`?

wpopielarski: done

PR #1155, opened 2017-04-12
Guards Scala Debug against fetching data outside of presentation comp…

wpopielarski:
…iler. Scala Model gets data eagerly.

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/core/internal/jdt/model/ScalaElements.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/core/internal/jdt/model/ScalaElements.scala
@@ -81,6 +81,7 @@ class ScalaSourceTypeElement(parent: JavaElement, name: String, declaringType: O
 
   override def getFullyQualifiedName: String =
     declaringType.flatMap { typ =>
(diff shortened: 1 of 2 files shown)


org.scala-ide.sdt.core/src/org/scalaide/core/internal/jdt/model/ScalaElements.scala
```diff
@@ -81,42 +80,44 @@ class ScalaSourceTypeElement(parent: JavaElement, name: String, declaringType: O
...
+    declaringType.flatMap { typ =>
+      pc.asyncExec {
+        val pkgSym = typ.typeSymbol.enclosingPackage
+        if (pkgSym.isEmptyPackage)
+          super.getFullyQualifiedName
+        else {
+          val pkg = pkgSym.javaClassName
+          pkg + "." + getTypeQualifiedName('$', /*showParameters =*/ false)
+        }
+      }.get.left.toOption
```
sschaef: If you do a `import org.scalaide.core.compiler.IScalaPresentationCompiler.Implicits._` you can do `.getOption()`, which has the advantage that a timeout will be used.

wpopielarski:
done, thanks

2017-04-13 12:48 GMT+02:00 Simon Schäfer <notifications@github.com>:

PR #1165, opened 2017-05-19
Enables using global Compiler Additional Parameters

wpopielarski:
Scenario:
- create new project, compiler uses default global additional params
- next change project's scala installation
- it should lead to situation where global settings are preceived and Xsource
and Xmacro-expand are added
- but if project specific setting's of additional params is not empty then switching scala installation should retain these settings and change Xsource and Xmacro-expand only

### DIFF
--- a/org.scala-ide.sdt.core/src/org/scalaide/core/internal/project/InstallationManagement.scala
+++ b/org.scala-ide.sdt.core/src/org/scalaide/core/internal/project/InstallationManagement.scala
@@ -1,24 +1,26 @@
 package org.scalaide.core.internal.project
 
-import org.scalaide.core.internal.jdt.util.ClasspathContainerSetter
-import org.eclipse.jface.preference.IPreferenceStore
-import org.scalaide.util.internal.CompilerUtils
-import scala.util.Try
-import org.scalaide.ui.internal.preferences.CompilerSettings
-import org.eclipse.jface.util.IPropertyChangeListener
-import org.scalaide.util.internal.SettingConverterUtil
-import org.scalaide.core.IScalaPlugin
-import org.eclipse.core.resources.IMarker
+import java.util.concurrent.atomic.AtomicReference
+
 import scala.tools.nsc.settings.ScalaVersion
 import scala.util.Failure
-import org.eclipse.jface.util.PropertyChangeEvent
 import scala.util.Success
-import org.scalaide.core.resources.MarkerFactory
+import scala.util.Try
+
+import org.eclipse.core.resources.IMarker
 import org.eclipse.core.runtime.Path
+import org.eclipse.jface.preference.IPreferenceStore
+import org.eclipse.jface.util.IPropertyChangeListener
+import org.eclipse.jface.util.PropertyChangeEvent
 import org.scalaide.core.IScalaPlugin
 import org.scalaide.core.ScalaInstallationChange
 import org.scalaide.core.SdtConstants
 import org.scalaide.core.internal.compiler.ScalaPresentationCompiler
+import org.scalaide.core.internal.jdt.util.ClasspathContainerSetter
+import org.scalaide.core.resources.MarkerFactory
+import org.scalaide.ui.internal.preferences.CompilerSettings
+import org.scalaide.util.internal.CompilerUtils
+import org.scalaide.util.internal.SettingConverterUtil
 
 trait InstallationManagement { this: ScalaProject =>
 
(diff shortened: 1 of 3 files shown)


org.scala-ide.sdt.core/src/org/scalaide/core/internal/project/InstallationManagement.scala
```diff
@@ -158,13 +160,22 @@ trait InstallationManagement { this: ScalaProject =>
     projectSpecificStorage.addPropertyChangeListener(compilerSettingsListener)
   }
 
+  private lazy val actualAdditionalParams: AtomicReference[() => String] = new AtomicReference(() => {
+    actualAdditionalParams.getAndSet(() => storage.getString(CompilerSettings.ADDITIONAL_PARAMS))
```
MelvinWM: Could `getAndSet` be replaced with `set`?

wpopielarski: not sure maybe it is too protective but idea is to set it atomic. `set` API does not guarantee it.

MelvinWM: There is actually some documentation regarding thread safety for `set`, though it is placed in the package documentation: http://docs.oracle.com/javase/8/docs/api/java/util/concurrent/atomic/package-summary.html.

wpopielarski:
looks it isn't
see:
https://github.com/openjdk-mirror/jdk7u-jdk/blob/master/src/share/classes/java/util/concurrent/atomic/AtomicReference.java#L91
and
https://github.com/openjdk-mirror/jdk7u-jdk/blob/master/src/share/classes/java/util/concurrent/atomic/AtomicReference.java#L139
key guy is `compareAndSet`

2017-06-05 17:28 GMT+02:00 MelvinWM <notifications@github.com>:

MelvinWM:
Ah, but `value` in that implementation has been declared volatile, so a single write to it (not a combined operation involving multiple steps like `x++`) is atomic and OK in regards to the JMM. I recall the book [Java: Concurrency in Practice](http://jcip.net/), which is a great book as I remember it and which also explains a bit about the Java Memory Model and the like, though it is somewhat old by now, and there regrettably hasn't been any new editions.

wpopielarski:
well, so looks like your understanding of concurrency in jvm is better than
mine :). actually I rather prefer to err on safe side than face some
unexpected and unpredictible behavior.  so maybe the code is too defensive
in this sense but I believe it doesn't have too big impact on performance.

2017-06-07 18:01 GMT+02:00 MelvinWM <notifications@github.com>:

rkrzewski:
@wpopielarski what's the status of this PR? It was sitting on the back burner for a long while :)

wpopielarski:
needs to be implemented, but implementation in this PR is incorrect.
Scenario is following:
1. there are some global settings
2. there are some project settings created from global settings (as simple
copy)
3. when I run project compilation then project settings if exists otherwise
global (current implementation)
4. so some guys want to have possibility to merge dynamically global
settings with project ones. So when I run project compilation then take
global settings and then adapt settings with project settings and use it as
current compilation settings. The biggest and unsolvable problem is
overriding settings which I want to remove from global with project
specific when flag exists or not. Example: there is -Xfatal-warnings flag
set in global and in project I want to suppress this flag but I cannot
because suppression means that I do not add this flag to project's
additional options, but on other hand it means that I get global settings
and this flag is active.

Hope I'm clear

2017-08-30 11:27 GMT+02:00 Rafał Krzewski <notifications@github.com>:

rkrzewski:
This looks like a potentially confusing departure from Eclipse model where "Use project settings" is is an all-or-nothing affair. Personally I would be against implementing this feature at all.

On the other hand, I also have a problem with `-Xfatal-warnings` I find it very disruptive during development in IDE. However I would much prefer to solve it on `sbteclipse` plugin's side: there should be a way to apply a transformation on the compiler settings that will be used when generating Eclipse projects, along the lines of
```
(scalacOptions in Eclipse) -= "-Xfatal-warnings"
```
I think it would be much more robust.

wpopielarski:
so you know now why this guy is not merged :)

2017-08-30 12:29 GMT+02:00 Rafał Krzewski <notifications@github.com>:

rkrzewski:
Yeah, thanks for the explanations. What about discarding this PR and filing an issue in `sbteclipse`? I could take a swing on implementing it. Or is there another use case except having `-Xfatal-warnings` on in SBT build but not in IDE that I'm missing?

wpopielarski:
According to the result of discussion on this functionality we decided to close this PR without no implementation.

wpopielarski:
I closed this PR. I think you are right with opening some issue/PR in
sbteclipse.

2017-08-30 13:53 GMT+02:00 Rafał Krzewski <notifications@github.com>:

rkrzewski:
:+1: I'll do that.

PR #1166, opened 2017-05-23
Enable projection document

wpopielarski:
Allows other bundles to set correct master document for its projection
in moment of its creation. This work is linked to problem with editting
of Play2 templates but can be used for any editor and structured
document.


sschaef:
LGTM

PR #1167, opened 2017-05-24
Include JDK version into compiler bridge cache path

rkrzewski:


sschaef:
I would also prefer @wpopielarski's schema of `java/scala/zinc` order

rkrzewski:
PR amended. I've changed the order. Are the prefixes OK? Also, I've used full JDK version eg 1.8.0_131 this means some potentially unnecessary rebuilds, but should be more resilient against potential changes in JDK9 early access releases.

PR #1175, opened 2017-08-16
Version checks for compiler plugins consistent with publisher's policy

rkrzewski:
Certain compiler plugins like and https://github.com/non/kind-projector
https://github.com/typelevel/machinist are published with cross version
corresponding to Scala binary version (eg. 2.12) others, like
https://github.com/scalamacros/paradise use full Scala version (eg.
2.12.3)

This change makes the check performed by IDE consistent with plugin
authors policy: plugin published with binary cross version will be
considered valid for any binary compatible Scala version, while plugins
published with full cross version will be considered valid only for
identical Scala version.

### DIFF
--- a/org.scala-ide.sdt.core.tests/src/org/scalaide/core/classpath/ClasspathTests.scala
+++ b/org.scala-ide.sdt.core.tests/src/org/scalaide/core/classpath/ClasspathTests.scala
@@ -505,8 +505,7 @@ class ClasspathTests {
   def errorKeptAfterClean(): Unit = {
     setRawClasspathAndCheckMarkers(cleanRawClasspath, 0, 1)
 
-    project.underlying.build(IncrementalProjectBuilder.CLEAN_BUILD, new NullProgressMonitor)
-    project.underlying.build(IncrementalProjectBuilder.FULL_BUILD, new NullProgressMonitor)
+    rebuild()
 
     checkMarkers(0, 1)
   }
@@ -524,8 +523,7 @@ class ClasspathTests {
     // illegal option
     projectStore.setValue(CompilerSettings.ADDITIONAL_PARAMS, "-Xi_dont_know")
 
-    project.underlying.build(IncrementalProjectBuilder.CLEAN_BUILD, new NullProgressMonitor)
-    project.underlying.build(IncrementalProjectBuilder.FULL_BUILD, new NullProgressMonitor)
+    rebuild()
 
     project.classpathHasChanged() // trick to make the check happen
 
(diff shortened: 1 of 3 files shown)


sschaef:
This change is not correct. Compiler plugins need to be published with the full Scala version because the compiler doesn't guarantee binary compatibility. Therefore, publishing a compiler plugin just with the binary version is wrong. This need to be fixed in the release process of the plugin - if we soften up the check in the IDE we only risk runtime crashes.

rkrzewski:
I'm sad to hear that. None of the plugins used by Typelevel projects (listed https://typelevel.github.io/projects/ at the bottom of the page) are published with full Scala version. I would much rather have relaxed requirements in the IDE, especially considering that even in case a compiler crash occurs it does not take whole IDE down, than to evangelize all plugin authors that their release process is wrong, or keep editing config files by hand (the status quo). I've been using a single version of kind-projector plugin with Scala 2.12.1 - 2.12.3 with no ill effects. A compiler crash would be an obvious indicator that it's time to look for updated plugin version.
 
@sschaef, perhaps the relaxed check could be optional, under a flag in Compiler Preferences panel? I could redo the PR in this fashion.

sschaef:
I can't remember which error is shown right now when the PC crashes due to binary incompatibilities, probably just some "PC crash" in the problems view. How many compiler plugins are there in the Typelevel project? Changing the publishing policy for a sbt project should be a one line fix (`crossVersion := CrossVersion.full`) but on the other side, changing the IDE is not difficult and I guess people who use compiler plugins know about potential problems with them.

rkrzewski:
I've never seen a compiler crash due to plugin incompatibility so I can't really tell. Right now classpath verification problem is reported and compiler refuses to start.

There are 7 compiler plugins under Typelevel umbrela but each one is maintained by different person.

I'm not sure what you mean by "changing the IDE is not difficult" here. Editing auto-generated configs by hand and creating symlinks is not difficult but it is tedious and repetitive. It bothered me enough to prepare this PR ;)

sschaef:
well ok, I'm going to merge this. If something breaks, we can still revert the changes.

rkrzewski:
I've provided tests to ensure that old behavior is preserved for plugins published with full versions. The new behavior is automating what I was doing by hand, so I believe it's an improvement overall.

In any case we can make this behavior optional with strict check by default.

sschaef:
Nah, I'm merging as is. People wouldn't understand such a check box that makes everything optional anyway.

rkrzewski:
Thanks!

PR #1177, opened 2017-08-21
Fix warnings

sschaef:
This enables some warnings that were enabled in the past but were disabled because no one wanted to fix them.


org.scala-ide.sdt.core/src/org/scalaide/refactoring/internal/method/ui/ChangeParameterOrderConfigurationPageGenerator.scala
```diff
@@ -81,7 +80,7 @@ trait ChangeParameterOrderConfigurationPageGenerator extends MethodSignatureRefa
         paramsWithSeparators: List[ParamOrSeparator]):
           List[ParamOrSeparator] = paramsWithSeparators match {
       case Nil => Nil
-      case p::Nil => paramsWithSeparators
+      case List(_) => paramsWithSeparators
```
wpopielarski: for me OK but mr Odersky would rather write it as `_ :: Nil`

PR #1183, opened 2017-10-28
upgrade to scalariform 0.2.6

godenji:


sschaef:
`0.2.5` is not correctly published. We need to wait for `0.2.6`

PR #1190, opened 2018-05-13
Scala 2.12.6 (WIP)

rkrzewski:
Update to Scala 2.12.6

All core tests pass
There are 2 failures in tests for debug expression evaluator that need to be investigated


m50d:
Is there a corresponding PR to scala-refactoring? I tried to build this (think I'm going to need an up-to-date 2.12 for JDK9 compatibility) but it seems to depend on an unpublished `org.scala-refactoring:org.scala-refactoring.library_2.12.6:jar:0.14.0-SNAPSHOT`.

rkrzewski:
You can use this branch https://github.com/rkrzewski/scala-refactoring/tree/scala-2.12.6
If my memory serves me, after building it locally with SBT, you'll need then use http://maven.apache.org/plugins/maven-deploy-plugin/deploy-file-mojo.html to publish the jar from your local ivy2 repository to your local m2 repository.

PR #1191, opened 2020-03-01
Updated path for Nebula Repo

chiragsanghavi:
As Nebula Repo has changed URL, its path needs to be updated.


chiragsanghavi:
Merge submitted by @rkrzewski is much more exhaustive then mine and already covers change that I have suggested. 

It would help if original committers of the project review and make this IDE relevant again.

rkrzewski:
@chiragsanghavi you should try https://scalameta.org/metals/ it works really well.

chiragsanghavi:
@rkrzewski  will try Metals. I saw it as a plug-in on VSCode. I wonder if Scala IDE is going to be revived at all.

rkrzewski:
No Scala IDE is dead for good. VirtusLab, the company that took over Scala IDE development from Lightbend is working on Metals nowadays.

Compared to Scala IDE, Metals is still missing a few bits of functionality (organize imports, most refactorings) but other that that is works faster and more reliably than Scala IDE ever did.

Metals started out as VSCode plugin, but now it supports multiple editors. There is also an Metals plugin for Eclipse - I haven't tried it though.

